PR #7, opened 2015-01-18
correcting transponed QR code

runn1ng:
Some QR scanners (javascript one, the app "QR Code Reader" or something like that on Android) have problems with transponed QR code. Trezor is generating transponed QR => I am correcting that

(it is not a nice fix but I am not sure how exactly the painting functions work to investigate further, also it's not a critical code so it doesn't really matter that much)


prusnak:
No need to investigate paint functions. Fix is OK.

runn1ng:
just to be sure for all the redditors lurking, I agreed with the relicencing

PR #16, opened 2015-02-22
PIN handling - constant time.

jhoenicke:
This diff contains three changes.
1. Make timing isPinCorrect independent of storage.pin, to avoid timing attacks
2. Only update failed PIN counter if the user entered a PIN.
   Of course, the fail counter is still incremented, before the PIN is checked.
3. Don't cache the PIN, but just the fact that the PIN was entered.  The
   cache should be in sync with storage.pin in any case.


prusnak:
Thanks!

PR #17, opened 2015-02-25
Faster signing, smoother progressbar.

jhoenicke:
Most time in signing transaction on the Trezor side is spent
in layoutProgress.  This patch reduces the calls to this functions.
We also compute the progress differently, reserving 50 % for downloading
input transactions and 50 % for the signing process.  This gives a
smoother experience if the input transactions are large.

If you use python-trezor you should apply the following diff, which increases the polling frequency.
```
diff --git a/trezorlib/transport_hid.py b/trezorlib/transport_hid.py
--- a/trezorlib/transport_hid.py
+++ b/trezorlib/transport_hid.py
@@ -122,7 +122,7 @@ class HidTransport(Transport):
                     # device is still alive
                     raise ConnectionError("Connection failed")
 
-                time.sleep(0.05)
+                time.sleep(0.0001)
                 continue
 
             report_id = data[0]
```

### DIFF
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -51,9 +51,14 @@ static uint8_t hash[32], hash_check[32], privkey[32], pubkey[33], sig[64];
 static uint64_t to_spend, spending, change_spend;
 const uint32_t version = 1;
 const uint32_t lock_time = 0;
-static uint32_t progress, progress_total;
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks! Much appreciated.

PR #23, opened 2015-03-29
Darkcoin to Dash re-branding

ELM4Ever:


prusnak:
I hope there won't be any more rebrandings in the following days. :D

PR #27, opened 2015-04-29
New usbDelay that delays and handles USB requests

jhoenicke:
Added usbDelay that polls usb port (for system requests) while delaying. This is called instead of delay in the button and pin delay functions. Experimental evaluation gave that the cycle count should be roughly divided by 28.5.

This patch fixes the problem where TREZOR gets unresponsive after python-trezor was used and there was a long delay between button presses.

### DIFF
--- a/firmware/protect.c
+++ b/firmware/protect.c
@@ -57,7 +57,7 @@ bool protectButton(ButtonRequestType type, bool confirm_only)
 
 		// button acked - check buttons
 		if (acked) {
-			delay(100000);
+			usbDelay(3500);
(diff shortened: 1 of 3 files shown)


jhoenicke:
One remaining problem is that during a oledRefresh, swipe, the wake up phase, etc, TREZOR is still unresponsive. But this can only be fixed by implementing interrupts handlers...

One also needs to recheck the delay when `libopencm3` is updated.  The timing of usbd_poll was evaluated experimentally.

prusnak:
Thanks!

PR #29, opened 2015-05-05
fixed salt passing to pbkdf2_hmac_sha512

dstanchfield:
Call to pbkdf2_hmac_sha512() uses salt string pointer argument (“TREZORHD”), which is located in ROM memory.  But pbkdf2_hmac_sha512() is expecting the pointer from RAM as it modifies the content. Actual salt value being used ends up being 
“TREZORHD\0Sta” as the attempt to modify it fails, (also producing a flash error flag to be set).


dstanchfield:
I'd also add, this effects users who may have used this feature (importing a bip-32 node and using a passphrase). So simply using this fix presents other issues of backwards compatibility.

prusnak:
Thankfully this option is not presented in myTREZOR interface and could have been only used via trezorctl command-line interface.

PR #34, opened 2015-07-04
fsm: simplify and refactor fsm_msgSignIdentity()

romanz:
unify SSH and regular signature logic.


romanz:
will rebase correctly and resubmit

romanz:
see #35

PR #35, opened 2015-07-04
fsm: simplify and refactor fsm_msgSignIdentity()

romanz:
this should minimize the different code between [SLIP-0013](http://doc.satoshilabs.com/slips/slip-0013.html) and SSH-related code.

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -653,6 +653,8 @@ void fsm_msgVerifyMessage(VerifyMessage *msg)
 	layoutHome();
 }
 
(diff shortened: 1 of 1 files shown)


romanz:
this code should work with the latest trezor-agent implementation (soon to be v0.3)

prusnak:
Already reworked in a different manner, but forgot to push 5f309641a164a4ca9af137a23ac1923a286270ed. Sorry. 

(Also there is a change in SLIP-0013 in a way how message is constructed. Will update that soon.)

romanz:
100%.
Thanks!

PR #42, opened 2015-08-24
firmware: exclude debug functionality from release build

romanz:
These functions are called only by `fsm_msgDebugLinkGetState()`, which is excluded from release build.


prusnak:
Nonreachable functions are AFAIK not included in the binary. But it's OK to mark them explicitly.

PR #43, opened 2015-09-06
bootloader: use regular object files

romanz:
`*.small.o` is the same as `*.o`


prusnak:
They are not the same on my system:

```
$ ls -l ecdsa*.o secp256k1*.o
-rw-r--r-- 1 stick users 91168 Sep  6 17:04 ecdsa.o
-rw-r--r-- 1 stick users 84952 Sep  6 17:04 ecdsa.small.o
-rw-r--r-- 1 stick users 39952 Sep  6 17:04 secp256k1.o
-rw-r--r-- 1 stick users  3036 Sep  6 17:04 secp256k1.small.o
```

romanz:
Oops... you are right.
They use different `CFLAGS` (`USE_PRECOMPUTED_IV` & `USE_PRECOMPUTED_CP`).
My mistake :(

PR #48, opened 2015-11-02
Add U2F support

darkskiez:
This adds a valuable new feature to the trezor, U2F Authentication support.

This isn't quite perfect yet, I'd like to do a little more to do with the interactions if both interfaces are attempted to be used simultaneously, possibly just disabling the u2f part when in usb_tiny mode, but I think it works pretty well as present, at least for testing, and getting some feedback.

I have successfully registered and authenticated to a few sites with this code, though not github yet as the settings are not visible on my account compared to their U2F FAQ.


prusnak:
Wow! Thanks! I will test this as soon as possible.

mikehearn:
:+1: wanna get a beer tonight mark? ;)

prusnak:
Lol, I want beer as well. :-)

darkskiez:
Is the submodule updated in master yet? :)

prusnak:
I pulled your changes into u2f* branch (for the time being).

* https://github.com/trezor/trezor-mcu/tree/u2f

btchip:
I'd suggest encrypting the App ID for privacy reasons (concern is an attacker collecting tokens and deducing which websites the user is browsing through cleartext hashes). For a sample implementation you can have a look at https://github.com/LedgerHQ/keepkey-firmware/blob/u2f/keepkey/local/baremetal/u2f_crypto.c

Bonus : you can also try to make the keyhandle smaller

Extra extended bonus with cherry on top : end up with a keyhandle + certificate small enough to save one HID request

darkskiez:
Cool! glad to see another implementation! 

I'm not sure I understand your concern though.. Attacker collects what tokens? 

In my implementation, the appid isnt in the keyhandle, it's random for every registration request. 

You can register to the same website and you will get a different keyhandle/pubkey each time.

Only the original trezor [the one with the right seed] can tell that the keyhandle is one that it produced, which is required by the u2f spec.

I can configure my code to also work with absolutely any size of keyhandle from 0 bytes to the max size, 

I went with 64 bytes for the keyhandle:

120 bits of entropy, 8 bits of convenience, and 128 bits of signature so the keyhandles were the same size as other tokens out there.

prusnak:
Argh, I really hate how Keepkey guys forked our software making it much more harder to incorporate their changes back. :-1: And now we ended up with 2 completely different U2F implementations. :-(

Also, @btchip do you want to share what is the relation between Ledger and Keepkey now that you started maintaining part of their codebase or is that a secret?

btchip:
@darkskiez oops my mistake, you're right, I read it too quickly. It's fine as it is, sorry. 

@prusnak no worries, my implementation is very easy to move from one project to another, and yours is fine too. No specific relations between Ledger and Keepkey, it was just the first thing I had in my bag for a quick hack and demoing protocol extensions.

Main differences, mine is slightly more complicated to read (designed to be used as a generic service and a hardware specific layer. It's intended to be used for a BLE nRF51 reference implementation when the specification goes public), and includes some extensions (using the screen to display information about the request, such as the site name or a picture).

Of course no preference if you pick one or the other.

darkskiez:
Ahh, that explains why it looks like it does :) It's quite pretty code.

You look like you're not generating bip32 hardened keys like I was earlier too!

I wrote this patch initially hoping to display the URL on the screen, but the current U2F spec doesn't allow it. I hope they add something optional in later releases. I have considered extending my implementation a bit and passing the appId to the client for it to hash itself as another message, but that would require writing a browser plugin with this feature.  Currently a quick look at your implementation requires 3rd party plugin support already to work so you could provide this feature too, perhaps more easily if you have the plugin extended already. I decided to make my implementation compatible at the u2f hid hardware level instead and work with chromes native support.

Your implementation also requires device unlocking.. As someone who uses one of these devices, multiple times every day, there is no way I could handle that. I would support making it optional in the device settings though. Currently as my implementation is driverless there is not an easy way to request the pin if it is needed.

btchip:
Mine also works with the default protocol, provided that you unlock the device before (either through a third party plugin or third party client). 

I think there''s an interesting design decision to discuss - users that put a PIN on their device might want to always have to enter that PIN whenever they're interacting with the device or get scared, and it's kind of compliant with the default U2F (the device could just prompt the user to unlock when a U2F request is received if still locked)

The extension-naming thing also works automagically for well known URLs (https://github.com/LedgerHQ/keepkey-firmware/blob/u2f/keepkey_board/local/baremetal/resources.c#L1916), with the option to add some dynamically through a third party plugin as you mentioned.

darkskiez:
I'm glad you've done that with the URLs, I did want to do exactly that too if it was suitable to put that sort of thing in the trezor storage.  My other idea was to take the url at time of registration, through third party plugin, pass it to the device with a vendor extension, then get the device to store it into the keyhandle when the user accepted it. The next time we went to that site, it would give us the data we needed over the normal protocols :)

First stop was getting a functioning implementation too.

prusnak:
@darkskiez Could you share how you generated the attestation keys? Thanks!

prusnak:
Jochen updated your efforts in his branch: https://github.com/jhoenicke/trezor-mcu/tree/u2f

We'll merge this stuff soon.

darkskiez:
Amazing, I'll have a look and include the notes for what I used to generate
those keys, tho they are totally useless unless you embed them in closed
hardware.

darkskiez:
From my notes i generated the keys and self signed cert with this:

openssl ecparam -genkey -out trezordevkey.pem -name prime256v1
openssl ec -in trezordevkey.pem  -text 
openssl req -new -key trezordevkey.pem -out trezordevcert.req -subj "/CN=Trezor U2F" 
openssl x509 -req -in trezordevcert.req -signkey trezordevkey.pem -days 3650 -out trezordevcert.pem
openssl x509 -in trezordevcert.pem  -C

I may have had to poke at openssl config to stop it adding all sorts of things by default..

jhoenicke:
I added an updated pull request #83.

prusnak:
Deprecated by https://github.com/trezor/trezor-mcu/pull/83

PR #72, opened 2016-02-18
bootloader autobuild with docker

karek314:
Added bootloader autobuild with docker and binary size checker


prusnak:
Could you rebase this to contain just one commit?

PR #74, opened 2016-02-18
Added bootloader autobuild with docker

karek314:
Added bootloader autobuild with docker and bootloader.bin size check

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -9,9 +9,11 @@ OBJDUMP = $(PREFIX)objdump
 FLASH   = st-flash
 OPENOCD = openocd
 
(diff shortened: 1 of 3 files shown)


karek314:
32000 bytes means max allowed bootloader size in case this mcu to work correctly.

prusnak:
It is 32768 actually.

karek314:
Actually it doesn't matter much since depending on settings in Makefile, bootloader was usually way above 32kb. But i have corrected. It may be useful when compiling bootloader first time.

prusnak:
Rebased commits into one da027f7652dd25ff7071562b57f3dd9a9349db72 and merged in. Thanks!

PR #78, opened 2016-04-20
Multicurve

jhoenicke:
Use different keys for different curves.

This requires the corresponding pull request in trezor-crypto and vice versa.


prusnak:
Thanks!

PR #79, opened 2016-04-20
Add GPG v2.1 support by signing message digest

romanz:
This should allow GPG signature creation with TREZOR.
See https://github.com/romanz/trezor-agent/issues/3 for more details.


romanz:
The Python code for generating GPG public keys and signatures is here:
https://github.com/romanz/trezor-agent/blob/gpg/gpg/signer.py#L82

prusnak:
Thanks!

PR #80, opened 2016-04-21
Remove the public_key hack.

jhoenicke:
It is no longer necessary to move the public key into a temporary buffer
since the node is specific for the curve and contains the right public
key.


prusnak:
Thanks!

PR #82, opened 2016-04-27
Don't reflash storage after each PIN entry

jhoenicke:
Instead of reflashing the whole storage, we use a designated area
in the second storage block, where we mark each PIN failure by a
single zero bit. This is because one can set bits in flash to zero but
not to one.  If the PIN was entered successfully the whole word is
set to zero and the next word stores the new PIN failure counter.


prusnak:
Tested with signed firmware and import from old storage works. Thanks!

PR #83, opened 2016-04-29
Updated u2f branch to current firmware

jhoenicke:
changes with regards to the previous pull request:
- generate keys on the nist256p tree.
- use smaller static buffers, moved attestation key to flash
    (the firmware was low on data memory)
- check for buffer overflows in u2fhid_read
- script to generate attestation key and certificate
- recognize well-known appids and display name (idea from btchip)

comments and peer reviews welcome :)


btchip:
Nice work, didn't review or test but as a quick note you should test against https://github.com/google/u2f-ref-code/tree/master/u2f-tests/HID which is more strict on timeouts and other protocol oddities. Once done it should be good to go.

jhoenicke:
Please don't merge this yet.  I think there is a stack overflow in it when it is jammed with messages:

`usbd_poll` -> `hid_u2f_rx_callback` -> `u2fhid_read` -> `u2fhid_msg` -> `LayoutHomeAfterTimeout` -> `usbPoll` -> `usbd_poll`

We have to redesign the communication loop to avoid this.  I have some ideas that should also help to pass the unit tests in `u2f-ref-code`.  But it means a bit of rewrite.

prusnak:
Ok, I will wait.

PR #86, opened 2016-05-24
Bootloader update: Faster flashing, smoother animation.

jhoenicke:
Compute sha256 hashsum at the end.
Display progress bar for flash erase.


prusnak:
Thanks! :-)

Issue #67, opened 2015-12-22
Initialize while waiting after wrong PIN


runn1ng:
When TREZOR is waiting for PIN matrix after repeated wrong PIN attempts, and computer stops communication, starts another communication and sends Initialize, TREZOR replies with PIN request and starts counting down again, even when Initialize shouldn't get PIN request.

Also, maybe more importantly, even when the computer correctly sends the PIN, the countdown starts *again*, as if the previous request didn't run.

Examples, using python-trezor and forking

* https://gist.github.com/runn1ng/c29dfc1e3c778c7a94bb
* https://gist.github.com/runn1ng/172ad85a5b6613f2b435

jhoenicke:
I analyzed it a bit.  One problem is that usbTiny is not set while waiting.  The init request is processed as usual but recursively inside the pin function, which is bad.  

I think the cleanest fix is to set usbTiny in the wait loop and when init is received abort the loop and return false (also call fsm_sendFailure which will answer the init request).

prusnak:
Fixed in #88

PR #88, opened 2016-05-28
Allow initialize() to abort PIN wait

jhoenicke:
See issue #67.

If the countdown for PIN failures is running and an initialize comes in, abort the countdown and cancel PIN entry.


jhoenicke:
Hmm, arm compiler is still broken in Travis.
```
The command "sudo apt-get install -y build-essential git gcc-arm-none-eabi" failed
```

prusnak:
Thanks! Will take a look at Travis build failure.

PR #89, opened 2016-06-06
u2f - add Slush Pool to WellKnown

ondrejsika:


prusnak:
https://slushpool.com/static/security/u2f.json returns error 500.

ondrejsika:
Yes, it has not implemented yet, but the json will be there.

Issue #85, opened 2016-05-16
Add ECDH support for GPG public-key decryption


romanz:
As an experimental patch, https://github.com/trezor/trezor-mcu/commit/ecfdc0dd29f020d9fcdfe95a70483b4f5961efcf seems to be working with NIST256P curve :)
In order to test it, I have added the following (very hackish - and soon-to-be-rewritten) Python code to the [ECDH branch](https://github.com/romanz/trezor-agent/compare/ecdh):

 - export GPG subkey supporting encryption: https://github.com/romanz/trezor-agent/commit/968c8bafccd5072e9bd3d2be0b9ea3e954776452
 - use "gpg-agent" emulation for public-key decryption: https://github.com/romanz/trezor-agent/commit/11ccc86618572548dc34267ad24bdac915abfa42

Also, I guess it would be nicer to add a separate API for this operation (since it's not really related to the "SignIdentity" API... I have added it there only to make the prototyping easier).

romanz:
Demo usage:

```

$ export TREZOR_GPG_USER_ID="testing"

$ trezor-gpg create -s > /tmp/key 
2016-05-16 21:22:16,897 INFO       nist256p1 GPG public key <8D17941D1414BD9C> created at 2016-05-16 21:22:13 for "testing"
2016-05-16 21:22:17,030 INFO       adding subkey to primary GPG key "testing" (A49CBFEE317EDBEF)
2016-05-16 21:22:17,030 INFO       confirm signing subkey with gpg-agent
2016-05-16 21:22:17,178 INFO       signing digest: D7459D69D2B6DD47BABA8AA4B2A6AB955094E3A0BBC00E3F1F790F65B57B3277

$ gpg2 --import /tmp/key 
gpg: key 317EDBEF: "testing" 1 new signature
gpg: key 317EDBEF: "testing" 1 new subkey
gpg: Total number processed: 1
gpg:            new subkeys: 1
gpg:         new signatures: 1

$ gpg2 -k
/home/roman/.gnupg/pubring.kbx
------------------------------
pub   nistp256/317EDBEF 2016-05-14 [SC]
uid         [ultimate] testing
sub   nistp256/3ECC4691 2016-05-16 [E]


$ python -m trezor_agent.gpg.agent &
[1] 6788

$ 2016-05-16 21:22:51,916 DEBUG      connected to Trezor F83618F42686D73F8D964F8F                                                         factory.py:27
[...]

PR #90, opened 2016-06-10
crypto: add ECDH session key generation

romanz:
This will allow to implement ECDH support for GPG 2.1+, after https://github.com/trezor/trezor-common/pull/12 is merged.
See https://github.com/trezor/trezor-mcu/issues/85 for more details.


prusnak:
Thanks!

PR #91, opened 2016-06-11
Update protobuf definitions

romanz:
The following commit will add a new handler for the `GetECDHSessionKey` request (#85).


prusnak:
Thanks!

PR #93, opened 2016-06-16
Add ECDH support to TREZOR API

romanz:
I was not sure about the correct way to confirm the ECDH session key generation (see the comment at line 787), so currently it uses PIN-based protection only, similar to the `GetPublicKey` API.

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -775,6 +775,54 @@ void fsm_msgSignIdentity(SignIdentity *msg)
 	layoutHome();
 }
 
+void fsm_msgGetECDHSessionKey(GetECDHSessionKey *msg)
+{
+	RESP_INIT(ECDHSessionKey);
+
+	if (!storage_isInitialized()) {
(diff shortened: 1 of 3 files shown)


romanz:
This should implement [SLIP-0017](https://github.com/satoshilabs/slips/blob/master/slip-0017.md), as described here: https://github.com/trezor/trezor-mcu/issues/85

prusnak:
I guess same protection as we use for GetPublicKey (i.e. ask for PIN) is fine. Thanks!

romanz:
Great, thanks!

jhoenicke:
Sorry for being a bit late on this.  Shouldn't there be a message similar to when signing identity.  Something like "Decrypt message for gpg://myidentity"?

For which other purposes than decryption do we expect ECDH to be used?

romanz:
It's possible to do so, but I wasn't sure about the actual layout & UX considerations, so I left it out for now...
@prusnak what do you think?

PR #97, opened 2016-06-29
Only compute pubkey on demand.

jhoenicke:
This is the second half of pull request trezor/trezor-crypto#61.  This changes the firmware to compute the public key where necessary and compute the fingerprint of the parent public key as required to produces BIP32 compatible xpub keys.

I'm not so happy with the way ed25519 public keys are handled.  I think they should be returned as 32 byte keys, but I didn't want to change the API without discussion and it would also break unit tests in python-trezor.  For backward compatibility I now replace the new 01 prefix in these public keys back to 00 again.

### DIFF
--- a/firmware/crypto.c
+++ b/firmware/crypto.c
@@ -85,13 +85,13 @@ uint32_t deser_length(const uint8_t *in, uint32_t *out)
 	return 1 + 8;
 }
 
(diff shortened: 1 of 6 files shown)


jhoenicke:
The check that failed is a timeout in test_codepoints.  Is there an easy way to increase the timeout limit for one test?

prusnak:
Merged in. Thanks!

PR #99, opened 2016-07-14
Ask for confirmation on ECDH Session (gpg decrypt)

jhoenicke:
Also fix the message "abort" to send a reply (an ActionCancelled failure)


prusnak:
Thanks!

PR #100, opened 2016-07-23
add python-ecdsa dependency

schinzelh:
I wanted to use `make sign` inside the container and found this dependency to be missing.


prusnak:
Thanks!

PR #103, opened 2016-08-17
Implement Ethereum support

axic:
This is the original tree rebased approx. 3 weeks ago. Cherry picked all recent changes.


axic:
I think it would be a good idea adding safety checks for transactions which probably the user didn't intended to do. There should be a way to disable these checks.

Examples:
- sending value to address 0
- sending transaction to address 0 without a data field (ie. an invalid deployment)

prusnak:
Checks implemented in https://github.com/trezor/trezor-mcu/commit/d0b12256954fbe4278f812a962331b3a386b9172

Feel free to add more if needed.

prusnak:
Checks implemented in https://github.com/trezor/trezor-mcu/commit/793eac06a04b5efd386c61aa1a97302f0d616a13

Let's not add the disable checks for now unless there are particular examples when this is necessary.

axic:
@prusnak added you as a contributor so you can directly commit to this PR

firmware/ethereum.c
```diff
@@ -280,6 +280,31 @@ static void layoutEthereumConfirmTx(const uint8_t *to, const uint8_t *value, uin
...
+		for (size_t i = 0; i < msg->to.size; i++) {
+			if (msg->to.bytes[i] > 0) {
+				address_zero = false;
+				break;
+			}
+		}
+	}
+
+	// sending value to address 0
+	if (address_zero && msg->has_value && msg->value.size) {
```
axic: Sorry I've realised that sending value is valid, but only part of contract deployment (initial stipend).

Basically the following check alone should be enough.

We could also check for sending data to a non-zero address which is most likely a contract invocation.We could warn the user it is not a regular transfer.

axic:
Perhaps displaying the first 4 bytes of the data sent to a non-zero address would also be useful so the user sees what contract invocation is being done.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+			fsm_sendFailure(FailureType_Failure_Other, "Data length provided, but no initial chunk");
+			ethereum_signing_abort();
+			return;
+		}
+		if (msg->data_initial_chunk.size > msg->data_length) {
+			fsm_sendFailure(FailureType_Failure_Other, "Invalid size of initial chunk");
+			ethereum_signing_abort();
+			return;
+		}
+		data_total = msg->data_length;
```
jhoenicke: There should be a check for data_length < 16000000 here, to prevent overflows later and the total rlp length from exceeding 2^24.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+}
+
+/*
+ * Push an RLP encoded length to the hash buffer.
+ */
+static void hash_rlp_length(uint32_t length, uint8_t firstbyte)
+{
+	uint8_t buf[4];
+	if (length == 1 && firstbyte == 0x00) {
+		// special case: null is encoded differently
```
jhoenicke: According to https://github.com/ethereum/wiki/wiki/RLP,  0x80 is the encoding of the empty string, not of the string containing a single 0 byte.

We shouldn't have the workaround here, instead I guess the higher layer should set value to the empty instead of "0", when a zero value  should be transferred.  Otherwise we cannot send a message with a single 0 byte.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+		buf[0] = 0x80;
+		hash_data(buf, 1);
+	} else if (length == 1 && firstbyte <= 0x7f) {
```
axic: This could be changed to `firstbyte > 0 && firstbyte <= 0x7f`. In that case the `firstbyte = 0 ` case would fall to the next condition and we could remove the special case on the top.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+
+	/* Stage 2: Store header fields */
+	hash_rlp_list_length(rlp_length);
+
+	layoutProgress("Signing", 100);
+
+	if (msg->has_nonce) {
+		hash_rlp_field(msg->nonce.bytes, msg->nonce.size);
+	} else {
+		hash_rlp_length(1, 0);
```
jhoenicke: This should be rlp_length(0, 0),  empty field: length is 0.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	if (msg->has_value) {
+		hash_rlp_field(msg->value.bytes, msg->value.size);
+	} else {
+		hash_rlp_length(1, 0);
+	}
+
+	if (msg->has_data_length && msg->has_data_initial_chunk) {
+		hash_rlp_length(msg->data_length, msg->data_initial_chunk.bytes[0]);
```
jhoenicke: This should use the value data_total instead of msg->data-length.  Also the if is unnecessary; if data_total is 0 (and therefore, also data_initial_chunk_size) the then part does the same as the else part.

firmware/ethereum.c
axic: And that should allow us to remove the workaround in `rlp_length`.

firmware/ethereum.c
```diff
@@ -0,0 +1,449 @@
...
+			return;
+		}
+		if (msg->data_initial_chunk.size > msg->data_length) {
+			fsm_sendFailure(FailureType_Failure_Other, "Invalid size of initial chunk");
+			ethereum_signing_abort();
+			return;
+		}
+		data_total = msg->data_length;
+	} else {
+		data_total = 0;
```
jhoenicke: Either check that has_data_initial_chunk is empty, or set data_total to its length (depending on what protocol you prefer).

jhoenicke:
There is a compilation problem.
s/DIALOG_ICON_QUESTION/&bmp_icon_question/;

jhoenicke:
Added simple confirmation dialog for raw data (only when given).

jhoenicke:
Field encoding needs more work: python-trezor sends address with leading zeros to Trezor but doesn't put leading zeros in raw transactions.  Also sending to empty address (new contract) doesn't  work.

jhoenicke:
There should be a fee estimate shown on Trezor.  I'm thinking of adding a third screen "Send xxxx ETH (including up to yyyy ETH fee)"

firmware/ethereum.c
```diff
@@ -0,0 +1,551 @@
...
+
+	layoutProgress("Signing", 100);
+
+	hash_rlp_field(msg->nonce.bytes, msg->nonce.size);
+	hash_rlp_field(msg->gas_price.bytes, msg->gas_price.size);
+	hash_rlp_field(msg->gas_limit.bytes, msg->gas_limit.size);
+	hash_rlp_field(msg->to.bytes, msg->to.size);
+	hash_rlp_field(msg->value.bytes, msg->value.size);
+	hash_rlp_length(data_total, msg->data_initial_chunk.bytes[0]);
+	if (data_total > 1 || msg->data_initial_chunk.bytes[0] >= 0x80) {
```
axic: `hash_rlp_length` could return the length or above the value of the appropriate `rlp_calculate_length` could be saved (` rlp_length += rlp_calculate_length(data_total, msg->data_initial_chunk.bytes[0]);`) and reused for this check:

`if (data_total > 1 || length > 1)`

However that is a minor improvement.

Can you please add a comment to explain why this condition is here? (So it won't get overlooked later on).

firmware/ethereum.c
```diff
@@ -0,0 +1,551 @@
...
+		}
+		/* Our encoding only supports transactions up to 2^24 bytes.  To
+		 * prevent exceeding the limit we use a stricter limit on data length.
+		 */
+		if (msg->data_length > 16000000)  {
+			fsm_sendFailure(FailureType_Failure_Other, "Data length exceeds limit");
+			ethereum_signing_abort();
+			return;
+		}
+		data_total = msg->data_length;
```
axic: Since `data_length` contains the entire length, `data_total` can be removed.

firmware/ethereum.c
```diff
@@ -0,0 +1,551 @@
...
+	} else {
+		data_total = 0;
+	}
+	if (msg->data_initial_chunk.size > data_total) {
```
axic: This could be folded into the above condition block.

firmware/ethereum.c
```diff
@@ -0,0 +1,551 @@
...
+	hash_rlp_field(msg->to.bytes, msg->to.size);
+	hash_rlp_field(msg->value.bytes, msg->value.size);
+	hash_rlp_length(data_total, msg->data_initial_chunk.bytes[0]);
+	if (data_total > 1 || msg->data_initial_chunk.bytes[0] >= 0x80) {
+		hash_data(msg->data_initial_chunk.bytes, msg->data_initial_chunk.size);
+	}
+	data_left = data_total - msg->data_initial_chunk.size;
+
+	/* FIXME: probably this shouldn't be done here, but at a later stage */
+	memcpy(privkey, node->private_key, 32);
```
axic: Is this still the best place to do this?

firmware/ethereum.c
prusnak: Not really. `msg->data_length` is available only in `ethereum_signing_init` function, while `data_total` is used in `send_request_chunk` which is outside of the visibility scope.

firmware/fsm.c
```diff
@@ -628,6 +653,47 @@ void fsm_msgGetAddress(GetAddress *msg)
...
+		return;
+
+	if (msg->has_show_display && msg->show_display) {
+		char desc[16];
+		strlcpy(desc, "Address:", sizeof(desc));
+
+		char address[41];
+		data2hex(resp->address.bytes, 20, address);
+
+		layoutAddress(address, desc);
```
axic: Is this definitely not truncating the address? It was a while back.

firmware/ethereum.c
axic: Ah right, for the progress display.

firmware/fsm.c
prusnak: Yes, changed in https://github.com/trezor/trezor-mcu/pull/103/commits/48008ddd8eeae19005011dc01a3d9bc24c1be244

firmware/ethereum.c
jhoenicke: I'm wondering if it is better if hash_rlp_length hashes nothing for length==1 and initial_byte < 0x80.  Then we can remove the `if` here and in hash_rlp_field.

firmware/ethereum.c
jhoenicke: But we want it to be executed in both cases: If data_length is missing, check that no initial chunk was given, if data_length is given, check that initial chunk is not too large.

firmware/ethereum.c
jhoenicke: I don't see a better way.  We don't want to change protocol to give the node only in the last ack and it is easier to store the private key instead of the derivation path.  Also this way we check that the derivation path works (i.e. storage is initialized) before asking the user to sign the message.

firmware/ethereum.c
axic: Oh, that's correct. :+1:

firmware/ethereum.c
```diff
@@ -0,0 +1,545 @@
...
+	return true;
+}
+
+void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
+{
+	ethereum_signing = true;
+	sha3_256_Init(&keccak_ctx);
+
+	memset(&resp, 0, sizeof(EthereumTxRequest));
+	/* set fields to 0, to avoid conditions later */
```
axic: Please update the comment here which explains that the upper Protobuf layers should do this, but ensure it in the case they didn't.

axic:
The only last question we had is whether `nonce` as a protocol field should be mandatory even in the case it is `0`. Right now it isn't mandatory.

prusnak:
It is not advised to put required fields in protobuf. Protobuf v3 even doesn't support required fields at all.

Checking for nonce in `ethereum_signing_check` is enough.

prusnak:
Merged!

Thank you all for contribution! @axic @Arachnid @jhoenicke

bgok:
Congrats guys. Nice work!

prusnak:
@bgok Start Your Photocopiers! :-D

bgok:
@prusnak Already on it 😎

tgerring:
Great work, very excited to try out!

PR #104, opened 2016-08-19
Refactor Makefile and Dockerfile

spudowiar:
* Complete refactor of the `Makefile` with colours and Git-based `clean` recipe!
    - Only print command run if command fails, prefer to print **what** is being done, rather than **how**
    - Allow a more pragmatic `make firmware` (aliased to `make`) or `make bootloader`
    - Support adding more "applications" (like `demo`, `firmware` or `bootloader`) using `make APPLICATION=<name>` and a directory called `<name>` containing a `Makefile.include`
* Follow `Dockerfile` practices

Screenshot during build

PR #105, opened 2016-08-20
Bootloader-in-firmware support

spudowiar:
This adds bootloader-in-firmware support (known internally as `fastflash`) which allows faster firmware flashing by using the new, improved firmware, hence the name.

From the commit message
> Allows using new, improved bootloader for faster flashing by building
> into firmware.
> 
> Pass `FASTFLASH=1` to `make` to build into firmware. Activate by holding
> the left button at the unofficial firmware confirmation.
> 
> `bootloader/trezor.bin` will be linked for fastflash usage and won't run
> as standard TREZOR bootloader.

This is based on the concept of [`jhoenicke:bootloaderhack`](https://github.com/jhoenicke/trezor-mcu/tree/bootloaderhack) and the branch's base is [`spudowiar:refactor-makefile`](https://github.com/trezor/trezor-mcu/pull/104) and this patch is incredibly useful during development on a production TREZOR device.

/cc @jhoenicke


memory_app_fastflash.ld
```diff
@@ -0,0 +1,9 @@
+/* STM32F205RE - 512K Flash, 128K RAM */
+/* program starts at 0x2001200 */
+MEMORY
+{
+	rom (rx) : ORIGIN = 0x20012000, LENGTH = 40K
```
jhoenicke: Currently rom and ram are overlapping.  Should be either moved to 0x20016000 or extended to 56K.

spudowiar: I originally started at 64K, dropped it down to 48K, and instead of dropping down to 40K, I accidentally brought it up to 56K!

spudowiar:
Closing until #104 is merged

PR #107, opened 2016-08-26
Hold confirm button to lock TREZOR and display screensaver

### LINKED ISSUE #60: indicate lock / unlock state on TREZOR screen + hold confirm button to lock
* motivation: TREZOR's should display it's state securely, not rely on software, locking by holding confirm button for 1 second in idle state is 10x more convenient then locking through browers, mainly when the mytrezor other wallet is not shown. Also we don't want to rely on specific wallet implementation to do this right.
* optionally this feature could be turned on / off in settings
* optionally there could be two types (locked / unlocked) of screen savers sets by user not one

spudowiar:
Partially fixes #60

### DIFF
--- a/firmware/trezor.c
+++ b/firmware/trezor.c
@@ -28,12 +28,48 @@
 #include "layout2.h"
 #include "rng.h"
 
+#include "buttons.h"
+
 uint32_t __stack_chk_guard;
 
 void __attribute__((noreturn)) __stack_chk_fail(void)
 {
 	layoutDialog(&bmp_icon_error, NULL, NULL, NULL, "Stack smashing", "detected.", NULL, "Please unplug", "the device.", NULL);
 	for (;;) {} // loop forever
+}
+
+void lock_device(void)
+{
+	static bool lockButtonDown = false;
+
+	buttonUpdate();
+	if (button.YesDown >= 100000) {
+		/* Button held down for long enough, wait for button release */
+		lockButtonDown = true;
+	} else if (lockButtonDown && button.YesUp) {
+		/* Button released */
+		lockButtonDown = false;
+
+		layoutDialog(&bmp_icon_question, "Cancel", "Lock Device", NULL, "Do you really want to", "lock your TREZOR?", NULL, NULL, NULL, NULL);
+
+        usbTiny(1);
+		do {
+			usbDelay(3300);
+			buttonUpdate();
+		} while (!button.YesUp && !button.NoUp);
+        usbTiny(0);
+
+		if (button.YesUp) {
+			/* Lock the device */
+			session_clear(true);
+			layoutScreensaver();
+		} else {
+			layoutHome();
+		}
+	} else {
+		/* Button possibly released during usbPoll() */
+		lockButtonDown = false;
+	}
 }
 
 int main(void)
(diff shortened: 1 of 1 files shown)


spudowiar:
The wording may appear more ominous that it should, any suggestions?

firmware/trezor.c
```diff
@@ -36,6 +38,38 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
...
+
+	buttonUpdate();
+	if (button.YesDown >= 200000) {
+		/* Button held down for long enough, wait for button release */
+		lockButtonDown = true;
+	} else if (lockButtonDown && button.YesUp) {
+		/* Button released */
+		lockButtonDown = false;
+
+		layoutDialog(&bmp_icon_question, "Cancel", "Lock Device", NULL, "Do you really want to", "lock your TREZOR?", NULL, NULL, NULL, NULL);
```
jhoenicke: Do we really want to confirm this?
We could also display a message "locking TREZOR" and swipe that away after half a second.

firmware/trezor.c
```diff
@@ -36,6 +38,38 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
 	for (;;) {} // loop forever
 }
 
+void lock_device(void)
+{
+	static bool lockButtonDown = false;
+
+	buttonUpdate();
+	if (button.YesDown >= 200000) {
```
jhoenicke: Is Yes or No button better?

firmware/trezor.c
```diff
@@ -36,6 +38,38 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
...
+		/* Button held down for long enough, wait for button release */
+		lockButtonDown = true;
+	} else if (lockButtonDown && button.YesUp) {
+		/* Button released */
+		lockButtonDown = false;
+
+		layoutDialog(&bmp_icon_question, "Cancel", "Lock Device", NULL, "Do you really want to", "lock your TREZOR?", NULL, NULL, NULL, NULL);
+
+		do {
+			delay(100000);
```
jhoenicke: Should be usbDelay.

firmware/trezor.c
spudowiar: @prusnak https://github.com/trezor/trezor-mcu/issues/60 (emphasis mine)

> indicate lock / unlock state on TREZOR screen + hold **confirm button** to lock

firmware/trezor.c
slush0: I agree, confirmation is unnecessary for such action. Displaying just information as Jochen suggested would be far enough.

spudowiar: Well, my only concern is for people like me who tend to mess around with the buttons :smile: Have you tried it? It feels rather fluid with confirmation.

slush0: I must admit I didn't try it yet :-). However generally we want to avoid unnecessary button confirmations, just to teachn people that pressing button is something they should care about.

spudowiar: I dunno, it's a very short delay and a button press and that just _feels_ right to me :) Try it!

prusnak:
Merged manually in 0b51d060d87774126181058beefcfeccc5ae436c

Thanks!

PR #108, opened 2016-08-26
Faster SPI communication

jhoenicke:
Reopening after fixing the bug (thanks @spudowiar)

According to specs, the display can handle up to 10 MHz SPI traffic
(minimum cycle time is 100 ns). This patch sets the SPI port to
7.5 MHz (60 MHz clock with divisor 8).  In my tests the display works
even fine with SPI frequency at 60 MHz.  I also minimized the delays
(the spec doesn't require a long setup time) and use the registers to
check when transfer is done instead of using a fixed delay.  This is the 
procedure suggested by the STM32F2 manual.

This patch also changes the swipe function to swipe pixel by pixel.
Otherwise, the swipe would happen too fast to be seen.

This changes the display update time from 11 ms to 1.5 ms.

### DIFF
--- a/oled.c
+++ b/oled.c
@@ -82,11 +82,12 @@ static bool is_debug_mode = 0;
 inline void SPISend(uint32_t base, uint8_t *data, int len)
 {
 	int i;
-	delay(400);
+	delay(1);
 	for (i = 0; i < len; i++) {
 		spi_send(base, data[i]);
 	}
-	delay(800);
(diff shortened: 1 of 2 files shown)


spudowiar:
I'd suggest that next time you simply reopen the previous pull request (GitHub will see the amended commits)

prusnak:
Thanks!

PR #111, opened 2016-08-30
Update coins.c for Zcash mainnet

bitcartel:
Address prefix is 'T' (uppercase) so address byte is 65 (decimal).

It was not clear from the CoinType protobuf definition what the byte 0x18 or 0x19 before the signed message represented.  Since only Bitcoin uses 0x18 and all others use 0x19, this patch uses 0x19.  Please change where appropriate.

### DIFF
--- a/firmware/coins.c
+++ b/firmware/coins.c
@@ -28,6 +28,7 @@ const CoinType coins[COINS_COUNT] = {
 	{true, "Litecoin", true, "LTC",  true,  48, true,    1000000, true,   5, false, 0, false, 0, true, "\x19" "Litecoin Signed Message:\n"},
 	{true, "Dogecoin", true, "DOGE", true,  30, true, 1000000000, true,  22, false, 0, false, 0, true, "\x19" "Dogecoin Signed Message:\n"},
 	{true, "Dash",     true, "DASH", true,  76, true,     100000, true,  16, false, 0, false, 0, true, "\x19" "DarkCoin Signed Message:\n"},
+	{true, "Zcash",    true, "ZEC",  true,  65, true,    1000000, true,   5, false, 0, false, 0, true, "\x19" "Zcash Signed Message:\n"},
 };
 
 const CoinType *coinByShortcut(const char *shortcut)


slush0:
Considering you have working testnet and testnet's Insight, could you provide also parameters for it?

firmware/coins.c
```diff
@@ -28,6 +28,7 @@ const CoinType coins[COINS_COUNT] = {
 	{true, "Litecoin", true, "LTC",  true,  48, true,    1000000, true,   5, false, 0, false, 0, true, "\x19" "Litecoin Signed Message:\n"},
 	{true, "Dogecoin", true, "DOGE", true,  30, true, 1000000000, true,  22, false, 0, false, 0, true, "\x19" "Dogecoin Signed Message:\n"},
 	{true, "Dash",     true, "DASH", true,  76, true,     100000, true,  16, false, 0, false, 0, true, "\x19" "DarkCoin Signed Message:\n"},
+	{true, "Zcash",    true, "ZEC",  true,  65, true,    1000000, true,   5, false, 0, false, 0, true, "\x19" "Zcash Signed Message:\n"},
```
prusnak: What string do you use for signing? Is it really "Zcash Signed Message:\n", if yes then the character preceeding it should not be \x19 (as it is the string length).

slush0:
Okay, what I've found in zcash-insight fork, it looks like zcash testnet uses the same address version as bitcoin testnet, so addition for zcash testnet into trezor is not necessary. The only downside is that message signing on zcash testnet won't work (because of "Bitcoin Signed Message:" string).

daira:
https://github.com/zcash/zcash/issues/1371 is our ticket to update ``strMessageMagic`` (currently in the Zcash Beta 2 milestone).

daira:
The signing prefix has been updated: https://github.com/zcash/zcash/commit/0163f8faa29a14f8757e5dce930f4c2530065215

We also changed the address prefixes for transparent addresses. The new prefixes are implemented in https://github.com/zcash/zcash/pull/1477/commits/702721c1fbde449a02a91a08bf63c3d7a6a7900c ; note that they are two bytes rather than one byte.

The address prefixes used in v1.0.0-beta-2 will not change again before launch.

daira:
Also see section 5.6.1 of our [protocol spec](https://github.com/zcash/zips/blob/master/protocol/protocol.pdf).

PR #116, opened 2016-10-06
Reworked rfc6979 signing.

jhoenicke:
This is the second part of the rfc6979 signing patch.  It must be applied together with the corresponding patch in trezor-crypto.

New parameter is_canonical that allows for generating signatures that
have additional requirements.


prusnak:
Please don't include the submodule change in the commit.

prusnak:
Thanks!

PR #122, opened 2016-10-17
Add specific layout for GPG signature

romanz:
Following ECDH usage of layoutDecryptIdentity (which shows "GPG decrypt for:" when used for ECDH) this commit adds a specific case for layoutSignIdentity, showing "GPG sign for:", instead of "GPG login to:" (which is less appropriate in the GPG context).


prusnak:
Thanks!

romanz:
Wow, that was quick :)

PR #123, opened 2016-10-20
Support for txs with no inputs

runn1ng:
Some transactions in zcash can have no inputs (all "inputs" are joinsplits, that are in extra data here).

I have tried to add support for transactions without inputs.

*However* - I might have added too many `if`s - because only the *previous* transactions can have 0 inputs, not the transaction I am currently signing (as of right now).

Also I haven't tested this at all :) will do tomorrow

### DIFF
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -276,7 +276,11 @@ void signing_init(uint32_t _inputs_count, uint32_t _outputs_count, const CoinTyp
 
 	layoutProgressSwipe("Signing transaction", 0);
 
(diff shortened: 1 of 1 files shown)


prusnak:
Will be fixed in next commit.

daira:
Remember that a Zcash transaction is allowed to have zero inputs only if it's a v2 (or higher, in future) transaction, marking that it has the ``nJoinSplit`` and ``vJoinSplit`` fields).

PR #128, opened 2016-11-17
Use Cortex-M3 SysTick timers

saleemrashid:
SysTick timers provide efficient and incredibly accurate timings. This pull request removes `usbDelay(uint32_t cycles)` in favour of `usbSleep(uint32_t millis)` which is implemented with the aforementioned SysTick timers.

### DIFF
--- a/firmware/usb.c
+++ b/firmware/usb.c
@@ -27,6 +27,7 @@
 #include "u2f.h"
 #include "storage.h"
 #include "util.h"
+#include "timer.h"
 
 #define USB_INTERFACE_INDEX_MAIN 0
 #if DEBUG_LINK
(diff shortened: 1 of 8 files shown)


prusnak:
Nice! Thanks!

firmware/usb.c
```diff
@@ -426,9 +427,10 @@ char usbTiny(char set)
 	return old;
 }
 
-void usbDelay(int cycles)
+void usbSleep(uint32_t millis)
 {
-	while (cycles--) {
+	uint32_t end = system_millis + millis;
+	while (end > system_millis) {
```
jhoenicke: It's probably not that critical, but the timer runs over every 4 million seconds, i.e. roughly every 50 days (if someone keeps the Trezor plugged in that long).

To avoid this, use `(int32_t) (end - system_millis) > 0` or `(system_millis - start) < millis`

saleemrashid: I did calculate this but considered it non-critical, as you did, since churning out more OpenPGP code is currently my priority.

But a solution (like the ones you posted) would be nice :wink:

PR #129, opened 2016-11-18
timer: Fix non-critical integer overflow

saleemrashid:
Every 4294967295 milliseconds (2 ^ 32 - 1), `system_millis` will overflow.
This means that every 49.71 days, `system_millis` will reset to zero.
Comparisons like `system_millis < (system_millis + 1)` would fail if the
latter had overflown and the former had not.

This is non-critical because the worst case is that one second could be
skipped or the screen could lock early.

This poses no threat to the exponential backoff used for protection
against brute force.

Thanks to @jhoenicke for the provided solution :heart:

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -84,7 +84,7 @@ void layoutHome(void)
 	oledRefresh();
 
 	// Reset lock screen timeout
-	system_millis_lock = system_millis + SCREEN_TIMEOUT_MILLIS;
+	system_millis_lock_start = system_millis;
 }
 
 const char *str_amount(uint64_t amnt, const char *abbr, char *buf, int len)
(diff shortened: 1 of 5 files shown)


prusnak:
@jhoenicke ack?

jhoenicke:
Okay, I checked to make sure.  Casting uint32 to int32 is "implementation defined" and gcc defines it the way we want it.   In principle a different compiler may throw an implementation defined exception, though.

Since we use gcc it is okay, but for more compatibility something like
`(system_millis - start_millis) < delay`
(where delay is an unsigned integer or constant) is more portable.  The behaviour of unsigned ints is more standardized than signed ints (uint32_t operations are always modulo 2^32 on over-/underflow).

saleemrashid:
@jhoenicke Updated.

prusnak:
Thanks!

PR #131, opened 2016-11-26
Fix screen timeout

saleemrashid:
Well, I'm an idiot :blush:


prusnak:
Haha, I have missed it too.

PR #132, opened 2016-11-26
layout: split "First Last <first@last.com>" GPG user ID for readability

romanz:
Following https://github.com/romanz/trezor-agent/issues/80, it should help displaying long GPG user IDs.

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -305,10 +305,12 @@ void layoutSignIdentity(const IdentityType *identity, const char *challenge)
 	char row_hostport[64 + 6 + 1];
 	char row_user[64 + 8 + 1];
 
+	bool is_gpg = (strcmp(identity->proto, "gpg") == 0);
+
 	if (identity->has_proto && identity->proto[0]) {
 		if (strcmp(identity->proto, "https") == 0) {
 			strlcpy(row_proto, "Web sign in to:", sizeof(row_proto));
-		} else if (strcmp(identity->proto, "gpg") == 0) {
+		} else if (is_gpg) {
 			strlcpy(row_proto, "GPG sign for:", sizeof(row_proto));
 		} else {
 			strlcpy(row_proto, identity->proto, sizeof(row_proto));
@@ -335,6 +337,21 @@ void layoutSignIdentity(const IdentityType *identity, const char *challenge)
 		strlcat(row_user, identity->user, sizeof(row_user));
 	} else {
 		row_user[0] = 0;
+	}
+
+	if (is_gpg) {
+		// Split "First Last <first@last.com>" into 2 lines:
+		// "First Last"
+		// "first@last.com"
+		char *email_start = strchr(row_hostport, '<');
+		if (email_start) {
+			strlcpy(row_user, email_start + 1, sizeof(row_user));
+			*email_start = 0;
+			char *email_end = strchr(row_user, '>');
+			if (email_end) {
+				*email_end = 0;
+			}
+		}
 	}
 
 	layoutDialogSwipe(&bmp_icon_question, "Cancel", "Confirm",


saleemrashid:
`First Last <user@domain.tld>` is a commonly used format. Maybe this shouldn't be GPG specific?

romanz:
I am not sure, since other use-cases of the SignIdentity() API (e.g. Web or SSH login) are filling the [IdentityType](https://github.com/trezor/trezor-common/blob/master/protob/types.proto#L261) protobuf fields for a login attempt to a specific user+server combination.
In the GPG case, the user ID may be any string - so I originally used the `host` field (to show it in the first line in the confirmation dialog UI). 
OTOH, it's possible to refactor this string-splitting logic to a separate function, to make it accessible for other use-cases than GPG.

prusnak:
I am not very happy about misusing various fields such as host just to fit the formatting function. I will merge this PR but refactor the layout function completely.

What is the current usage of IdentityType for GPG?

Would the following be an option?

IdentityType.proto = gpg
IdentityType.user = first.last
IdentityType.host = server.com
IdentityType.port = unused
IdentityType.path = unused

Resulting in gpg://first.last@server.com URI

And adding an extra field to IdentityType called 'description' which will contain "First Last". (Should this be a part of IdentityFingerprint hash or not?)

saleemrashid:
From [RFC 4880][1]

> A User ID packet consists of UTF-8 text that is intended to represent
> the name and email address of the key holder.  By convention, it
> includes an [RFC 2822][2] [[RFC2822][2]] mail name-addr, but there are no
> restrictions on its content.  The packet length in the header
> specifies the length of the User ID.

[1]: https://tools.ietf.org/html/rfc4880#section-5.11
[2]: https://tools.ietf.org/html/rfc2822

prusnak:
I know the RFC. What I am talking about is whether it is possible to create a mapping from OpenPGP UserID to IdentityType, or whether we just want to give up and just use OpenPGP UserID as IdentityType.host field, which is completely against all conventions. (Also, do we need the reverse mapping, i.e. should we be able to loslessly reconstruct OpenGPG UserID from IdentityType structure? If that's not the case I would vote for the first option.)

romanz:
I agree with @prusnak that the current state of GPG's usage of SignIdentity with IdentityType.host containing the whole user ID is hacky indeed...
Maybe we could add a separate API for this operation?
This way, it would share the same HD derivatation logic with [SLIP13](https://github.com/satoshilabs/slips/blob/master/slip-0013.md#hd-structure), but allow any user ID as its "[URI](https://github.com/satoshilabs/slips/blob/master/slip-0013.md#service-identity)".
It would be possible for the user to split the user ID into several lines (using a `repeated string` field), which will be shown in separate lines during the confirmation dialog.
What do you think?

prusnak:
Can we first evaluate my questions from my comments above before introducing more concepts?

romanz:
Sorry for derailing the discussion...

I think that your [suggestion](https://github.com/trezor/trezor-mcu/pull/132#issuecomment-263541045) for adding a `description` field for holding the name (and also a possible comment) should work well.
The issue here is that GPG user ID is an unrestricted text field, so many different formats are used, e.g.:
```
pub   rsa4096 2014-11-19 [SC]
      C52261BE0A823221D94CA1D1CB11CA1D03FA5082
uid           [ unknown] Qubes OS Release 3 Signing Key

pub   rsa4096 2015-02-07 [SC]
      ED727C306E766BC85E621AA65FA6C3E4D9AFBB99
uid           [ unknown] Joanna Rutkowska (Master Signing Key)

pub   rsa4096 2011-06-15 [SC]
      6694D8DE7BE8EE5631BED9502BD5824B7F9470E6
uid           [ unknown] Thomas Voegtlin (https://electrum.org) <thomasv@electrum.org>
uid           [ unknown] ThomasV <thomasv1@gmx.de>
uid           [ unknown] Thomas Voegtlin <thomasv1@gmx.de>
sub   rsa4096 2011-06-15 [E]
```
I don't think this should be a problem for parsing user ID into IdentityType and vice-versa (as you [suggested](https://github.com/trezor/trezor-mcu/pull/132#issuecomment-263566073)), but unfortunately all the parts of the IdentityType structure should be hashed using IdentityFingerprint (since it's possible to specify a user ID without an email address).

To conclude, I am fine with any decision you make for the key derivation - just wanted to know how we should handle long user IDs (in the GPG case). As far as I'm concerned, it's fine to keep the current implementation for now, until TREZOR 2 is ready and we'll have a lot of screen real estate :)

prusnak:
Ah, that's bad it is possible to use UserID without username@host.com part :-/ I'll think about the options.

prusnak:
Ok, let's use this simple hack for now. I don't see any other simple way without breaking too much things.

prusnak:
Thanks!

PR #135, opened 2016-12-30
usb: exclude hid_report_descriptor_debug from non-debug build

romanz:
Found at https://travis-ci.org/trezor/trezor-mcu/builds/187775326#L957


saleemrashid:
@romanz I've noticed that's been like that for a while but assumed it'd be stripped out of the binary. Seems odd the error would just appear like that.

prusnak:
Thanks!

PR #138, opened 2017-01-10
memory protection: do not write reserved bits

mruddy:
Hi, based on the [hardware BOM](https://doc.satoshilabs.com/trezor-tech/hardware.html) that lists STM32F205RE as the MCU being used, I reviewed the bootloader protection code this morning.

I'm proposing some updates to not set reserved bits (which may lead to undefined behavior?) and to remove an unnecessary call to `flash_unlock_option_bytes` see also: https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/common/flash_common_f24.c#L402-L413

I also added references to the manual that explains more about what is going on. I have not tested this code, so, please review.

### DIFF
--- a/memory.c
+++ b/memory.c
@@ -27,13 +27,23 @@
 
 void memory_protect(void)
 {
+	// Reference STM32F205 Flash programming manual revision 5 http://www.st.com/resource/en/programming_manual/cd00233952.pdf
+	// Section 2.6 Option bytes
 	//                     set RDP level 2                   WRP for sectors 0 and 1
-	if ((((*OPTION_BYTES_1) & 0xFFFF) == 0xCCFF) && (((*OPTION_BYTES_2) & 0xFFFF) == 0xFFFC)) {
+	if ((((*OPTION_BYTES_1) & 0xFFEC) == 0xCCEC) && (((*OPTION_BYTES_2) & 0xFFF) == 0xFFC)) {
 		return; // already set up correctly - bail out
 	}
 	flash_unlock_option_bytes();
-	//                                WRP +    RDP
-	flash_program_option_bytes(0xFFFC0000 + 0xCCFF);
+	// Section 2.8.6 Flash option control register (FLASH_OPTCR)
+	//   Bits 31:28 Reserved, must be kept cleared.
+	//   Bits 27:16 nWRP: Not write protect: write protect bootloader code in flash main memory sectors 0 and 1 (Section 2.3; table 2)
+	//   Bits 15:8 RDP: Read protect: level 2 chip read protection active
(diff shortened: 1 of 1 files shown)


memory.c
```diff
@@ -27,13 +27,22 @@
 
 void memory_protect(void)
 {
+	// Reference STM32F205 Flash programming manual revision 5 http://www.st.com/resource/en/programming_manual/cd00233952.pdf
+	// Section 2.6 Option bytes
 	//                     set RDP level 2                   WRP for sectors 0 and 1
 	if ((((*OPTION_BYTES_1) & 0xFFFF) == 0xCCFF) && (((*OPTION_BYTES_2) & 0xFFFF) == 0xFFFC)) {
```
prusnak: Shouldn't values in these conditions be also changed? (to 0xCCEC and 0x0FFC, respectively?)

memory.c
```diff
@@ -27,13 +27,22 @@
...
 		return; // already set up correctly - bail out
 	}
-	flash_unlock_option_bytes();
```
prusnak: Good find, but I don't think keeping this has any side effects. I would like to keep that because it is more explicit.

memory.c
mruddy: No, I don't think so based on table 7 in section 2.6.1. It shows bit 4 of the user byte being 1 and bits 15-12 of the other as being 0xF. Do you read that table differently?

mruddy: Hmm, on second look, you might be right about bits 1:0 of the option bytes at 0x1FFF C000. The table I mentioned shows them as 0x1. So maybe that means OPTSTRT should be 0. That would make sense.

prusnak: Hm, maybe we can use different masks so we don't compare reserved bits. My point is that results in condition should be ideally same as values we set below, otherwise it is confusing.

mruddy: makes sense, i'll update the masks

memory.c
mruddy: OK, I'll put it back. My main reason for removing it was because of the process to modify the option bytes (section 2.6.2) having the first step of waiting for any pending flash operations to complete. The libopencm3 code does the unlock after waiting, but the manual doesn't say if anything bad/unspecified will happen if that register is changed while another operation is pending. I still kind of lean towards not having the extra unlock. But, it's hard to say if it matters (though probably not).

mruddy:
put up a new commit that is squashed and incorporates all feedback so far

mruddy:
are there any other nits etc... that you'd like changed, or is this just baking before merge?

prusnak:
Thanks!

PR #139, opened 2017-01-10
Check that U2F key only uses hardened derivation

jhoenicke:
We generate only U2F keys with hardened derivation.  However, we
didn't check incoming keys if they used hardened derivation.
This patch fixes this.


firmware/u2f.c
```diff
@@ -503,8 +503,15 @@ const HDNode *generateKeyHandle(const uint8_t app_id[], uint8_t key_handle[])
 const HDNode *validateKeyHandle(const uint8_t app_id[], const uint8_t key_handle[])
 {
 	uint32_t key_path[KEY_PATH_ENTRIES];
+	unsigned int i;
```
prusnak: We use C99 in most of the new code, so it is OK to declare loop variable in for cycle below.

jhoenicke: fixed.

prusnak:
Thanks!

PR #144, opened 2017-01-21
WebUSB: Deduplicate endpoint

saleemrashid:
Interfaces are virtually unlimited, using the same endpoint for both
interfaces frees up an extra endpoint.

This has the disadvantage that claiming the WebUSB interface, also
"claims" the HID interface. This is, however, a far more appealing
disadvantage than the loss of that endpoint altogether.

### DIFF
--- a/firmware/usb.c
+++ b/firmware/usb.c
@@ -46,8 +46,6 @@
 
 #define ENDPOINT_ADDRESS_DEBUG_IN   (0x82)
 #define ENDPOINT_ADDRESS_DEBUG_OUT  (0x02)
-#define ENDPOINT_ADDRESS_WEBUSB_IN     (0x82)
-#define ENDPOINT_ADDRESS_WEBUSB_OUT    (0x02)
 
 #define ENDPOINT_ADDRESS_U2F_IN     (0x83)
 #define ENDPOINT_ADDRESS_U2F_OUT    (0x03)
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks! Will rebase the branch afterwards.

prusnak:
I squashed your and Karel's changes into one commit, so it is now obvious what was changed against the original implementation. Please check the squashed commit in the `webusb-wip` branch.

// cc @runn1ng

saleemrashid:
@prusnak Could you squash this in

```patch
diff --git a/firmware/usb.c b/firmware/usb.c
index 88d8b21..24ad169 100644
--- a/firmware/usb.c
+++ b/firmware/usb.c
@@ -38,2 +38,3 @@
 #define USB_INTERFACE_INDEX_U2F 2
+#define USB_INTERFACE_INDEX_WEBUSB 3
 #else
@@ -309,3 +310,2 @@ static const struct usb_interface ifaces[] = {{
 	.altsetting = hid_iface_u2f,
-#if !(DEBUG_LINK)
 }, {
@@ -313,3 +313,2 @@ static const struct usb_interface ifaces[] = {{
 	.altsetting = hid_iface_webusb,
-#endif
 }};
@@ -445,3 +444,2 @@ void usbInit(void)
 	usb21_setup(usbd_dev, &bos_descriptor);
-#if !(DEBUG_LINK)
 	static const char* origin_urls[] = {
@@ -455,3 +453,2 @@ void usbInit(void)
 	webusb_setup(usbd_dev, origin_urls, sizeof(origin_urls)/sizeof(origin_urls[0]), USB_INTERFACE_INDEX_WEBUSB);
-#endif
 }
```

prusnak:
squashed

saleemrashid:
Just noticed that, that should be like before but with 4 and 3. Thanks.

prusnak:
Makes sense, fix force-pushed to `webusb-wip` branch again.

PR #147, opened 2017-01-30
Clean up USB strings

saleemrashid:
* Replace hard coded USB string identifiers with `enum`
* Annotate interfaces with human readable names


prusnak:
Thanks!

PR #148, opened 2017-01-31
Rubycoin added

Stoner19:


prusnak:
We don't add random coins into our firmware. Does your coin have working infrastructure? (i.e support in Electrum, working Bitcore/Insight fork, etc?)

Issue #150, opened 2017-02-14
using RNG peripheral before setup?


mruddy:
I was reviewing the RNG code alongside the [reference manual](http://www.st.com/resource/en/reference_manual/cd00225773.pdf) (the latest linked to from http://www.st.com/en/microcontrollers/stm32f205re.html), and I don't see how it's not the case that either the bootloader is relying on undocumented behavior, or the reference manual has some errors.

Since I'm not sure which is the case, I'm split on if this is an issue. My device seems to work, but I don't see a guarantee that this will continue to be the case, for example, for other production runs. Below's what I see. Please let me know what you think, or if I'm missing something.

1. Section 20.4.1 says that the RNG control register resets at 0x00000000 which should mean that the TRNG peripheral is disabled. This is in line with section 20.3.1 since step 2 activates the analog part. The way I read things, until the RNGEN bit is set, the part should not be active. And at reset, the bit is documented to be cleared.
2. For the stack guard, the bootloader attempts to read from the TRNG before calling the setup routine that enables the clock and sets the RNGEN bit in the control register. See https://github.com/trezor/trezor-mcu/blob/master/bootloader/bootloader.c#L142-L143 and https://github.com/trezor/trezor-mcu/blob/master/setup.c#L43-L44. The confusing thing is that the RNG_SR_DRDY bit must be getting set in the status register for random32 to read data, weird...
3. Also, for some reason related to FIPS, after setup is called, according to section 20.3.1, "the first random number generated after setting the RNGEN bit should not be used, but saved for comparison with the next generated random number".

mruddy:
@prusnak I made some quick custom firmware that prints the ```__stack_chk_guard``` just after the bootloader turns over control and the stack guard is always zero. So, this is actually looks like a valid bug.

saleemrashid:
> My device seems to work, but I don't see a guarantee that this will continue to be the case, for example, for other production runs.

Having a stack canary set to zero isn't going to stop it from working, but it is a security issue (reduces the stack smashing protection to protection against accidental smashing rather than malicious smashing).

> 3. Also, for some reason related to FIPS, after setup is called, according to section 20.3.1, "the first random number generated after setting the RNGEN bit should not be used, but saved for comparison with the next generated random number".

This is not too much of an issue since we probably do not need the canary to be cryptographically secure.

mruddy:
1. I agree that a constant canary could be a security issue.
[...]

PR #151, opened 2017-02-15
fix issue #150

mruddy:
Here's a patch to address #150. It needs review and testing, especially since it changes the bootloader and I'm coding while sleepy.

The only thing that may be surprising is that I removed the call to setup and oledInit at the beginning of the firmware. My reason for doing so is that the bootloader already handles those things.

My basic firmware tests of this patch passed, but I'll do some more testing tomorrow.

### DIFF
--- a/bootloader/bootloader.c
+++ b/bootloader/bootloader.c
@@ -139,8 +139,8 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
 
 int main(void)
 {
-	__stack_chk_guard = random32();
 	setup();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
 	memory_protect();
 	oledInit();
 
--- a/demo/demo.c
+++ b/demo/demo.c
@@ -248,11 +248,15 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
 
 int main(void)
 {
-	__stack_chk_guard = random32();
 #ifndef APPVER
 	setup();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
 	oledInit();
+#else
+	setupApp();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
 #endif
+
 	usbInit();
 
 	passlen = strlen((char *)pass);
--- a/firmware/trezor.c
+++ b/firmware/trezor.c
@@ -89,12 +89,13 @@ void check_lock_screen(void)
 
 int main(void)
 {
-	__stack_chk_guard = random32();
 #ifndef APPVER
 	setup();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
 	oledInit();
 #else
 	setupApp();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
 #endif
 
 	timer_init();
--- a/setup.c
+++ b/setup.c
@@ -21,6 +21,7 @@
 #include <libopencm3/stm32/gpio.h>
 #include <libopencm3/stm32/spi.h>
 #include <libopencm3/stm32/f2/rng.h>
+#include "rng.h"
 
 void setup(void)
 {
@@ -42,6 +43,9 @@ void setup(void)
 	// enable RNG
 	rcc_periph_clock_enable(RCC_RNG);
 	RNG_CR |= RNG_CR_IE | RNG_CR_RNGEN;
+	// to be extra careful and heed the STM32F205xx Reference manual, Section 20.3.1
+	// we don't use the first random number generated after setting the RNGEN bit in setup
+	random32();
 
 	// set GPIO for buttons
 	gpio_mode_setup(GPIOC, GPIO_MODE_INPUT, GPIO_PUPD_PULLUP, GPIO2 | GPIO5);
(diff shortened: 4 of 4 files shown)


mruddy:
pushed new commit. had to put oledInit back because it was init'ing static data. i tested more and have not encountered any problems.

what dev board do you guys use for testing the bootloader?

prusnak:
> what dev board do you guys use for testing the bootloader?

I would recommend Dinosaur Hiphop Zero: 

http://www.stellaw.info/blog/2015/12/16/breakout-board

or STM32F4DISCOVERY board.

mruddy:
@prusnak thanks for the code review and dev board info! updates made.
on my latest commit, i was wondering what you thought about setting the stack guard after oledInit?
putting before could delay oledInit if for some reason random32 takes perceptibly long. but, the tradeoff is that setting it as early as possible after setup gives the best stack smashing protection against future code updates. in the bootloader, i chose best security, in the other places, i went with slightly less and less code duplication.

prusnak:
> i was wondering what you thought about setting the stack guard after oledInit?

I think it is OK.  `oledInit` does not do anything with user specified input.

mruddy:
right on. update made. in effect, setup, setupApp, memory_protect, oledInit, oledClear, oledRefresh and anything they ever call are just in a separate security group that need to be well vetted against stack smashes in any future updates. that seems reasonable.

saleemrashid:
@mruddy But, at that stage, USB hasn't been initialized, so you're pretty safe against stack smashing. The real risk, as I see it, is that you perform stack smashing in the bootloader in order to bypass the unofficial firmware prompt.

bootloader/bootloader.c
```diff
@@ -139,11 +139,15 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
 
 int main(void)
 {
-	__stack_chk_guard = random32();
 	setup();
 	memory_protect();
 	oledInit();
 
+	// to be extra careful and heed the STM32F205xx Reference manual, Section 20.3.1
```
saleemrashid: Does this have to come after `memory_protect()` and `oledInit()`? If not, should this be moved to `setup()` and `setupApp()`?

bootloader/bootloader.c
```diff
@@ -139,11 +139,15 @@ void __attribute__((noreturn)) __stack_chk_fail(void)
...
+	// we don't use the first random number generated after setting the RNGEN bit in setup
+	random32();
+	__stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
```
saleemrashid: Does this have to come after `memory_protect()` and `oledInit()`? If not, should this be moved to `setup()` and `setupApp()`?

prusnak: I am not sure you can modify `__stack_chk_guard` variable outside of the `main()` function (because the function that sets `__stack_chk_guard` will break the check on return).

mruddy: the call to random32 that discards the result does not have to come after memory_protect and oledInit. calling it later only avoids any possibly (although unlikely) perceptible delay before init'ing the screen. this is what was discussed two comments above (i had both of these lines just after the call to setup). one could also argue that maybe we should not progress to init the screen until the rng is generating quickly/successfully. it may be useful to have the discarding random32 call in setup() and setupApp(). doing that makes sense to me. @prusnak is there a reason why the demo.c code does not, or could not, call setupApp near the beginning of its main function? if it was there this refactor would be complete.

changing the stack guard from within a function that uses it causes that function to fail on return though because the hidden value pushed on the stack (by the compiler generated code) won't match anymore. so the second line can't be factored into setup and setupApp.

prusnak: > is there a reason why the demo.c code does not, or could not, call setupApp near the beginning of its main function

No reason, you can add it if you want

mruddy:
@saleemrashid yep, i agree with your assessment of the stack smashing. i think it's important to have the stack protection in place simply at some point before we start processing usb and button inputs from outside the device. until that point though the vulnerability, in reasonable threat models that i can think of, is limited to self-inflicted/developer error (that i think can be reasonably mitigated with code review).

bootloader/bootloader.c
saleemrashid: @prusnak I think explicit modification of the canary might work, but I'm also not 100% sure.

saleemrashid: I remember there were some issues when I tried replacing parts of the firmware with Rust and there was some useful discussion on Gitter about it

mruddy:
refactored and also brought the ```__stack_chk_guard``` initialization back up to the earliest points before oled init etc...

bootloader/bootloader.c
mruddy: nah, can only modify the canary within main because it doesn't return. there's literally nowhere to even return to since ```reset_handler``` doesn't push anything onto the stack, it just branches. going into main, the MSP points just after the end of SRAM. ```load_app``` does the same thing when going from bootloader to firmware.

i verified that with my latest commit, and then adding the following patch, the stack smashing message displays:
```
diff --git a/setup.c b/setup.c
index 2537a01..e73fe5c 100644
--- a/setup.c
+++ b/setup.c
@@ -23,6 +23,8 @@
 #include <libopencm3/stm32/f2/rng.h>
 #include "rng.h"
 
+extern uint32_t __stack_chk_guard;
+
 void setup(void)
 {
        // setup clock
@@ -79,6 +81,7 @@ void setupApp(void)
        // "Each subsequent generated random number has to be compared with the previously generated
        // number. The test fails if any two compared numbers are equal (continuous random number generator test)."
        random32();
+       __stack_chk_guard = 1;
        // hotfix for old bootloader
        gpio_mode_setup(GPIOA, GPIO_MODE_INPUT, GPIO_PUPD_NONE, GPIO9);
        spi_init_master(SPI1, SPI_CR1_BAUDRATE_FPCLK_DIV_8, SPI_CR1_CPOL_CLK_TO_0_WHEN_IDLE, SPI_CR1_CPHA_CLK_TRANSITION_1, SPI_CR1_DFF_8BIT, SPI_CR1_MSBFIRST);
```

saleemrashid: @mruddy Can confirm

You can use my branch [`ssp`](https://github.com/saleemrashid/trezor-mcu/tree/ssp) to mess around with it, using the app in `ssp/`. It's compiled with `-O0`. If you call `hulkSmash()` to trigger a stack smash, it displays `Hulk smash!`, if you trigger a stack smash some other way, it displays `Unknown stack smash!` and if no stack smash is triggered, it displays `No stack smash...`

prusnak:
Commited as https://github.com/trezor/trezor-mcu/commit/c8c85424b629e4790f0745d15a1d4368046dff33

Thank you!

PR #153, opened 2017-02-21
possible rng error hang and recovery

mruddy:
first, this change does not set the interrupt enable bit in the rng control register. currently, the interrupt is not handled, it's not even enabled in the nvic.

second, this change checks the SECS and CECS bits in the rng status register directly, rather than indirectly via SEIS and CEIS.
the way i read the manual, SECS and SEIS, and CECS and CEIS, match when an error is first detected.
but, SEIS and CEIS do not stay synched with the clock and fault detector as SECS and CECS do.
so, when an error occurs SEIS and CEIS go high and stay high because our code never resets them.
it's possible that an error passes and SECS and CECS then go low again. that's why it's better to check them instead.
as it is, if an rng error occured, things would appear to just hang.

the other way to look at it is that once an rng error is detected, it's better to just unplug the device and totally start over.
we don't do any of the recommened error recovery -- even this PR just waits until the rng hardware hopefully self-corrects.
but, i think this PR gets closer to the sweet spot in the ability to recover without user interaction given the kinds of possible error conditions that i see being possible.

from the STM32F205xx reference manual section 20.4.2:

  Bit 6 SEIS: Seed error interrupt status:
    This bit is set at the same time as SECS, it is cleared by writing it to 0.

  Bit 5 CEIS: Clock error interrupt status:
    This bit is set at the same time as CECS, it is cleared by writing it to 0.

  Bit 2 SECS: Seed error current status:
    0: No faulty sequence has currently been detected. If the SEIS bit is set, this means that a faulty sequence was detected and the situation has been recovered.

  Bit 1 CECS: Clock error current status:
    0: The RNG_CLK clock has been correctly detected. If the CEIS bit is set, this means that a clock error was detected and the situation has been recovered

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -59,6 +59,7 @@ CFLAGS += -Wno-sequence-point
 CFLAGS += -Iprotob -DPB_FIELD_16BIT=1
 CFLAGS += -DQR_MAX_VERSION=0
 CFLAGS += -DDEBUG_LINK=0
+CFLAGS += -DDEBUG_RNG=0
 CFLAGS += -DDEBUG_LOG=0
 CFLAGS += -DSCM_REVISION='"$(shell git rev-parse HEAD | sed 's:\(..\):\\x\1:g')"'
 CFLAGS += -DED25519_CUSTOMRANDOM=1
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -292,12 +292,14 @@ void fsm_msgFirmwareUpload(FirmwareUpload *msg)
 
 void fsm_msgGetEntropy(GetEntropy *msg)
 {
+#if !DEBUG_RNG
 	layoutDialogSwipe(&bmp_icon_question, "Cancel", "Confirm", NULL, "Do you really want to", "send entropy?", NULL, NULL, NULL, NULL);
 	if (!protectButton(ButtonRequestType_ButtonRequest_ProtectCall, false)) {
 		fsm_sendFailure(FailureType_Failure_ActionCancelled, "Entropy cancelled");
 		layoutHome();
 		return;
 	}
+#endif
 	RESP_INIT(Entropy);
 	uint32_t len = msg->size;
 	if (len > 1024) {
--- a/firmware/trezor.h
+++ b/firmware/trezor.h
@@ -31,6 +31,10 @@
 #define DEBUG_LINK 0
 #endif
 
+#ifndef DEBUG_RNG
+#define DEBUG_RNG 0
+#endif
+
 #ifndef DEBUG_LOG
 #define DEBUG_LOG 0
 #endif
--- a/rng.c
+++ b/rng.c
@@ -27,7 +27,7 @@ uint32_t random32(void)
 {
 	static uint32_t last = 0, new = 0;
 	while (new == last) {
-		if (((RNG_SR & (RNG_SR_SEIS | RNG_SR_CEIS)) == 0) && ((RNG_SR & RNG_SR_DRDY) > 0)) {
+		if ((RNG_SR & (RNG_SR_SECS | RNG_SR_CECS | RNG_SR_DRDY)) == RNG_SR_DRDY) {
 			new = RNG_DR;
 		}
 	}
--- a/setup.c
+++ b/setup.c
@@ -42,7 +42,7 @@ void setup(void)
 
 	// enable RNG
 	rcc_periph_clock_enable(RCC_RNG);
-	RNG_CR |= RNG_CR_IE | RNG_CR_RNGEN;
+	RNG_CR |= RNG_CR_RNGEN;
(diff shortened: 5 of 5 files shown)


mruddy:
see also
https://github.com/trezor/python-trezor/pull/104
https://github.com/trezor/rng-test/pull/1

firmware/fsm.c
```diff
@@ -292,12 +292,14 @@ void fsm_msgFirmwareUpload(FirmwareUpload *msg)
 
 void fsm_msgGetEntropy(GetEntropy *msg)
 {
+#if DEBUG_LINK == 0
```
prusnak: What do you think about not using DEBUG_LINK here, but rather introducing another option, such as DEBUG_RNG?

saleemrashid: This should be `#if !DEBUG_LINK`

mruddy: sure, done

firmware/trezor.h
```diff
@@ -31,6 +31,10 @@
 #define DEBUG_LINK 0
 #endif
 
+#ifndef DEBUG_RNG
+#define DEBUG_RNG 0
```
saleemrashid: This is redundant, `#if` handles undefined macros

firmware/Makefile
```diff
@@ -59,6 +59,7 @@ CFLAGS += -Wno-sequence-point
 CFLAGS += -Iprotob -DPB_FIELD_16BIT=1
 CFLAGS += -DQR_MAX_VERSION=0
 CFLAGS += -DDEBUG_LINK=0
+CFLAGS += -DDEBUG_RNG=0
```
saleemrashid: See https://github.com/trezor/trezor-mcu/pull/153/commits/a69266651233df13b082806315919a614d12067f#r103516055

firmware/trezor.h
mruddy: i included this one to be consistent with DEBUG_LINK. i can remove it

firmware/Makefile
mruddy: yep, understood. just a matter of should i keep it consistent with how DEBUG_LINK is configured?

saleemrashid: I think that maybe all the definitions should be removed

/cc @prusnak

firmware/trezor.h
prusnak: That is true, unless `-Wundef` is used and that is exactly what we have in the `Makefile`.

saleemrashid: @prusnak Mea culpa, I was testing with `-Wall`, which doesn't include `-Wundef`

firmware/Makefile
prusnak: Please leave that in, see comment above why.

mruddy: ok

prusnak:
Btw, have you seen any changes in tests[1] after applying this change?

[1] https://github.com/trezor/rng-test

mruddy:
nothing consistent. i looked at diffs of the outputs vs the prior outputs and one test will be weak where another passed. between runs they vary too where one says weak and the next will say passed. the dieharder tests run so long that they're not done yet. if there are some specific measures that i should be more interested in, please let me know since i've never had to interpret any of these types of results before.

mruddy:
@prusnak i just ran a fresh set of entropy pulls and ran the "ent" and "rngtest"s and put the outputs here https://github.com/trezor/rng-test/compare/master...mruddy:test-run-output?expand=1

mruddy:
if this helps at all:

this branch has the ent and rngtest outputs from all of this PR's changes: https://github.com/mruddy/rng-test/tree/test-run-output

this branch has the ent and rngtest outputs from only this PR's DEBUG_RNG changes that make gathering the entropy feasible: https://github.com/mruddy/rng-test/tree/current-vs-original

diff
https://github.com/mruddy/rng-test/compare/mruddy:test-run-output...mruddy:current-vs-original

mruddy:
here's a full result set from another run: https://github.com/trezor/rng-test/compare/master...mruddy:full-results?expand=1

the files named like this are without this pull's rng.c and setup.c changes:
stm32_rng_1.dieharder, stm32_rng_1.ent, stm32_rng_1.rngtest

the files named like this include all of this pull's changes:
stm32_rng_1_pr153.dieharder, stm32_rng_1_pr153.ent, stm32_rng_1_pr153.rngtest

my unexpert interpretation is that this PR's changes don't look to make things worse. intuitively, that is my expectation since these are mechanical changes to use the peripheral in a slightly more straightforward way. these results may be more interesting compared to some changes like were pondered in https://github.com/trezor/trezor-mcu/issues/155. but, i haven't gotten to that yet.

mruddy:
in summary, this pull adds testing capability, avoids a potential hang, and makes it clear that an interrupt is not being used. in my opinion, it's ready to merge. thanks!

PR #158, opened 2017-03-29
Compute hash before checking signatures.

jhoenicke:
This fixes the problem where the wrong hash is shown, if the firmware contains no signing key indices.


prusnak:
Thanks!

PR #161, opened 2017-04-20
GetPublicKey altcoin support

peterzen:
The `GetPublicKey` call takes an additional `coin_name` parameter, looks up the `xpub_magic` value and and passes it on to `hdnode_serialize_public`.


prusnak:
Thanks!

PR #167, opened 2017-05-28
Generate MessagesMap in messages.c

saleemrashid:
This pull request adds `firmware/protob/messages_map.py` which generates `MessagesMap` from `firmware/protob/messages.proto`

It relies on the `wire_tiny` and `wire_bootloader` options from trezor/trezor-common#29


saleemrashid:
This doesn't produce the same build due to rearranged messages and additional types.

However, if you apply [this patch](https://github.com/trezor/trezor-mcu/files/1034455/reproduce-build-for-1.5.0.txt), you can get an identical build to https://beta-wallet.trezor.io/data/firmware/trezor-1.5.0.bin

```bash
git checkout messages-map
git reset --soft 6b74139b4530a4687b4a317b8b08f4329704efc4 # add 1.5.0 to changelog
git apply reproduce-build-for-1.5.0.txt # GitHub doesn't support .patch, grrr

```

jhoenicke:
There should be a "This file is automatically generated -- do not edit" comment at the beginning of `messages_map.h`.

prusnak:
1) Please use `// TINY` or something similar for lines that are commented because of `wire_tiny`.
2) Agree with the comment above
3) Please exclude `vendor/trezor-common` and `firmware/protob/types.pb.{c,h}` from PR.

saleemrashid:
> 1. Please use // TINY or something similar for lines that are commented because of wire_tiny.

I didn't do it because the existing code didn't have anything like that but I'll add that in.

> 3. Please exclude vendor/trezor-common and firmware/protob/types.pb.{c,h} from PR.

When you regenerate the Protocol Buffers, `firmware/protob/types.pb.[ch]` will be updated with the new extensions `wire_tiny` and `wire_bootloader`. Also, the script will not work without the updated `vendor/trezor-common` (as it won't be able to import `wire_tiny` and `wire_bootloader`

prusnak:
Thanks! Added small changes in https://github.com/trezor/trezor-mcu/commit/34c0b93d549b634b3e3cd4126a1229621bad19a8

saleemrashid:
@prusnak I had literally just finished working on some changes but your changes look much better :+1:

PR #172, opened 2017-06-10
Second bootloader support (fastflash)

saleemrashid:
Add second bootloader support to enable faster flashing of firmware. This is similar to @jhoenicke's [`bootloaderhack`](https://github.com/jhoenicke/trezor-mcu/tree/bootloaderhack), but has a few advantages

* It does not affect normal builds of the firmware or the bootloader (so it can be merged into `master`)
* ~~The linker script reserves exactly the size of the bootloader~~ (not compatible with GCC 4.9.3)
* The linker script avoids use of magic constants (more avoidance is possible but it breaks compatibility with GCC 4.9.3)
* The linker verifies that the bootloader fits into the reserved space
* The bootloader does not waste time verifying the signatures (and unconditionally wipes the metadata)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: false
 dist: trusty
 language: c
 
+env:
+  global:
+    - MAKEFLAGS=-j2
+
 addons:
   apt:
     packages:
@@ -13,7 +17,7 @@ install:
   - export PATH="$PWD/gcc-arm-none-eabi-4_9-2015q3/bin:$PATH"
 
 script:
-  - CFLAGS="-std=c99" make -C vendor/libopencm3
+  - CFLAGS="-std=c99" make -C vendor/libopencm3 lib/stm32/f2
   - make
   - make -C firmware
   - make -C bootloader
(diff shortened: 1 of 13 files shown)


saleemrashid:
@prusnak PING :)

saleemrashid:
@prusnak `PING`

jhoenicke:
I see that you reserve 20 kB of precious RAM for the bootloader, even if it is not loaded.  It looks like we have enough memory now, but it would be good to avoid that. 

BTW, do you know a good tool to compute the maximum stack usage?  The code is non-recursive and I think doesn't use dynamic array sizes, so it should be possible to compute this automatically.

saleemrashid:
@jhoenicke It's the safest way to do this because that linker feature is designed precisely for situations like this.

Furthermore, the likelihood of running out of RAM is miniscule (`.data` + `.bss` is 80 KiB), we're far more likely to run out of ROM (which happens with GCC 4.8.2).

saleemrashid:
@prusnak The last two commits are optional because #178 reduces the firmware size which enables compilation with GCC 4.8.2

prusnak:
Merged except last two commits. Thanks!

Issue #59, opened 2015-11-13
Test recovery with initialized device


prusnak:
- same as standard recovery from user's point of view
- device will not store the seed, but shows "Recovery seed checked." // "Recovery seed mismatch."

Feature to make sure, that the seed works. Will be used before wiping the device before sending it for RMA etc.

ondra-novak:
BUMP

Please add as "verify recovery" Currently I have device at risk  because I am not sure, that I have correct sheet with the words (and which one is correct, because I have more such sheets). This can be solved only by moving all funds to other wallet, wipe device, recreate the seed and move funds back to the Trezor.. With ability to verify recovery I can skip this annoying procedure.

prusnak:
Other option perform the check on another TREZOR device.

romanz:
What do you think about the following solution?

https://github.com/trezor/trezor-common/compare/master...romanz:dry-run
https://github.com/trezor/trezor-mcu/compare/master...romanz:dry-run
https://github.com/trezor/python-trezor/compare/master...romanz:dry-run

The changes above add a new command for running the new Matrix-based recovery process in "dry run" mode, only to see if the mnemonic is correct.

romanz:
P.S.
Feel free to ignore and re-generate the protobuf-related commits at trezor-mcu and python-trezor.

jnicolau:
Is this feature on the make. Currently there is no way to test the seed? Any workaround?

romanz:
@prusnak Ping :)
(let me know if it's not the intended way for implementing this feature...)

prusnak:
Will review it soon.

jhoenicke:
I thought about a different feature that would also allow to check the recovery seed.  Have only one new message that exchanges the current configuration with another one stored at an offset in the configuration sector.  It is like switching between two different Trezors.  This also allows for securely updating to a new seed without the need to recover twice.  And it has other use cases such as having one seed for private use and one seed for business use.

Your solution is in some sense simpler since it only does one thing.  On the other hand the solution with multiple configurations, despite being more versatile, doesn't require any changes to the recovery function. Only one additional call to switch between configurations (and the code that writes to the storage sector needs to be changed to preserve the second configuration).  So it would be less or about equal amount of code in the firmware.

[...]

PR #176, opened 2017-06-21
Allow testing recovery with an initialized device

romanz:
This would allow safe mnemonic validation by using a dry-run flag (following https://github.com/trezor/trezor-mcu/issues/59).

**NOTE**: this PR requires re-generating protobuf-related code after https://github.com/trezor/trezor-common/pull/33 is merged.

### DIFF
--- a/firmware/recovery.c
+++ b/firmware/recovery.c
@@ -44,6 +44,11 @@ static uint32_t word_count;
  */
 static int awaiting_word = 0;
 
+/* True if we should not write anything back to storage
+ * (can be used for testing seed for correctness).
+ */
+static bool dry_run;
+
 /* True if we should check that seed corresponds to bip39.
  */
 static bool enforce_wordlist;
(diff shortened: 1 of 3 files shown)


firmware/recovery.c
```diff
@@ -121,28 +128,60 @@ static void recovery_request(void) {
...
+				storage.has_imported = true;
+				storage.imported = true;
+			}
+			storage_commit();
+			fsm_sendSuccess("Device recovered");
+		} else {
+			// Inform the user about new mnemonic correctness (as well as whether it is the same as the current one).
+			const char *msg = is_same_mnemonic(new_mnemonic) ? "(same as current)" : "(differs from current)";
+			layoutDialog(&bmp_icon_ok, NULL, NULL, NULL, "The mnemonic is", "successfully verified.", msg, NULL, NULL, NULL);
+			usbSleep(RECOVERY_DRY_RUN_DELAY_MS);
```
jhoenicke: This means the dialog is shown for 3 s?  Shouldn't it be confirmed by a button press?

firmware/recovery.c
```diff
@@ -121,28 +128,60 @@ static void recovery_request(void) {
...
+			fsm_sendSuccess("Mnemonic is correct");
```
jhoenicke: Is it intentionally not telling over USB whether the mnemonic matches?

firmware/recovery.c
```diff
@@ -121,28 +128,60 @@ static void recovery_request(void) {
...
+				// not enforcing => mark storage as imported
+				storage.has_imported = true;
+				storage.imported = true;
+			}
+			storage_commit();
+			fsm_sendSuccess("Device recovered");
+		} else {
+			// Inform the user about new mnemonic correctness (as well as whether it is the same as the current one).
+			const char *msg = is_same_mnemonic(new_mnemonic) ? "(same as current)" : "(differs from current)";
+			layoutDialog(&bmp_icon_ok, NULL, NULL, NULL, "The mnemonic is", "successfully verified.", msg, NULL, NULL, NULL);
```
jhoenicke: The message

    The mnemonic is
    successfully verified.
    (differs from current)

could be easily misread.  Maybe not say successfully if the mnemonic differs?

romanz:
Thanks for the review, Jochen!
I'll be happy to fix the code, but I'm not sure about the exact wording of the messages...
@prusnak What do you think?

firmware/recovery.c
saleemrashid: Furthermore, whether or not it is valid doesn't really matter to the user at this stage (otherwise they would have been told it was invalid)

jhoenicke:
I tested it now.  The message at the end should definitely wait for a button press.  The first time I didn't see the screen in time, since my eyes were still focused on the computer.

firmware/recovery.c
prusnak: Yes, please used button press.

firmware/recovery.c
prusnak: Hm, I would suggest using the following message scheme:

Success() - mnemonic is OK and matches
Failure() - mnemonic is OK but does not match
Failure() - mnemonic is invalid

romanz:
Thanks @jhoenicke, @prusnak and @saleemrashid  for the review, the testing and the suggestions!
I'll fix the issues and update the PR in the following days.

jhoenicke:
BTW, dry-run also works with a wiped Trezor.  In that case the mnemonic never matches :)

firmware/recovery.c
romanz: Thanks for the suggestion - done at https://github.com/trezor/trezor-mcu/pull/176/commits/efc07a0436bba6c6be3aac92a166c33cd7b6173a.

firmware/recovery.c
romanz: Done (please see https://github.com/trezor/trezor-mcu/pull/176/commits/efc07a0436bba6c6be3aac92a166c33cd7b6173a).

xbach:
Instead of "mnemonic", I would prefer the usage of the word "seed". While it is not necessarily technically correct, the majority of users remember the 24 words as "Recovery Seed" or "Seed" and not "Mnemonic Seed". :)

"The seed is valid and matches the existing one."
"The seed is valid but does not match the existing one."
"The seed is invalid."

prusnak:
I am not sure about "the existing one" wording. To me it is not obvious that "existing one" equals to "the one in the device".

prusnak:
Anyways, code looks fine. I'll merge it and we'll deal with wording later. Thank Roman, good work!

romanz:
Thanks :)

saleemrashid:
I think it should be something like

1. The seed is not valid
2. The seed does not match the device
3. The seed matches the device

("the device" could be changed, "seed" could be "recovery seed" if there's space)

prusnak:
See https://github.com/trezor/trezor-mcu/commit/825342529348c55184ded9ff00bc934eaf501893 for the current wording.

PR #177, opened 2017-06-22
Set -mcpu=cortex-m3 linker flag.

jhoenicke:
This fixes a bug with Ubuntu 17.04.
It includes the correct architecture (armv7-m) and also includes the flag -mfix-cortex-m3-ldrd as default.


saleemrashid:
ACK

prusnak:
Thanks!

Bellmoney:
I am having trouble getting my ETH from coinbase with this device. It will not pickup their address. I sent money into limbo

PR #179, opened 2017-06-26
Deferred backup

prusnak:
We'd like to make backup workflow independent from initialization.

When ResetDevice message contains field `skip_backup=true`, it will not do the backup, but rather set a field `needs_backup=true` in Storage.

Later user can invoke backup using BackupDevice message which will setup `needs_backup` back to false to avoid multiple invocations of this workflow.

When ResetDevice does not contain this field, original behavior is expected.

@jhoenicke @saleemrashid could you please check whether I got everything allright?

### DIFF
--- a/firmware/reset.c
+++ b/firmware/reset.c
@@ -33,12 +33,14 @@
 static uint32_t strength;
 static uint8_t  int_entropy[32];
 static bool     awaiting_entropy = false;
+static bool     skip_backup = false;
 
-void reset_init(bool display_random, uint32_t _strength, bool passphrase_protection, bool pin_protection, const char *language, const char *label, uint32_t u2f_counter)
+void reset_init(bool display_random, uint32_t _strength, bool passphrase_protection, bool pin_protection, const char *language, const char *label, uint32_t u2f_counter, bool _skip_backup)
 {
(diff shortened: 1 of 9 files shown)


firmware/reset.c
```diff
@@ -92,6 +92,33 @@ void reset_entropy(const uint8_t *ext_entropy, uint32_t len)
...
+
+static char current_word[10], current_word_display[11];
+
+void reset_backup(void)
+{
+	if (!storage.has_needs_backup || !storage.needs_backup) {
+		fsm_sendFailure(FailureType_Failure_UnexpectedMessage, _("Seed already backed up"));
+		return;
+	} else {
+		storage.has_needs_backup = false;
```
saleemrashid: This makes more sense as `storage.has_needs_backup = true`

saleemrashid:
ACK a63abf2 but it would be nice to use something like `DEVICE NOT BACKED UP` as the homescreen until the device is backed up

saleemrashid:
@prusnak Should probably add `needs_backup` to `Features` as well

prusnak:
Something is broken with this PR at Github side. Will open a new one => https://github.com/trezor/trezor-mcu/pull/180

PR #180, opened 2017-06-26
Deferred backup

prusnak:
We'd like to make backup workflow independent from initialization.

When ResetDevice message contains field skip_backup=true, it will not do the backup, but rather set a field needs_backup=true in Storage.

Later user can invoke backup using BackupDevice message which will setup needs_backup back to false to avoid multiple invocations of this workflow.

When ResetDevice does not contain this field, original behavior is expected.

@jhoenicke @saleemrashid could you please check whether I got everything allright?

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -234,6 +234,7 @@ void fsm_msgGetFeatures(GetFeatures *msg)
 	resp->has_imported = true; resp->imported = storage.has_imported && storage.imported;
 	resp->has_pin_cached = true; resp->pin_cached = session_isPinCached();
 	resp->has_passphrase_cached = true; resp->passphrase_cached = session_isPassphraseCached();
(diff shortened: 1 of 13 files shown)


jhoenicke:
reset_backup can wipe the Trezor if the user presses the wrong button. Or am I reading this line wrong?
https://github.com/trezor/trezor-mcu/pull/180/files#diff-4dbd9c0f675e913051933a9056b22c1aR169

prusnak:
Added b5f42c1:

reset_backup now has parameter to indicate whether it was called as a part of reset workflow or a separate backup workflow

also storage is now marked as backed up only when backup was performed until the end

prusnak:
Rebased and merged to master in https://github.com/trezor/trezor-mcu/commit/8fa3ce7e1491bd79d47eff10ffac758fcaf5c45e

PR #182, opened 2017-06-30
recovery: zero new_mnemonic memory before returning to the user

romanz:
Note: I wasn't sure whether to use `MEMSET_BZERO` macro (from `trezor-crypto`) or just plain `memset` (since most of the `trezor-mcu` code uses it).


prusnak:
Thanks!

PR #184, opened 2017-06-30
add ability for developers to easily avoid locking their flash memory

mruddy:
This PR adds the ability for developers to easily compile a bootloader that won't lock the bootloader flash memory sectors on their dev board :)

The default is still to lock them. This is just a manual override.

verified the generated code with:
```
make clean && make && arm-none-eabi-objdump -d memory.o
make clean && make MEMORY_PROTECT=1 && arm-none-eabi-objdump -d memory.o
make clean && make MEMORY_PROTECT=0 && arm-none-eabi-objdump -d memory.o
```

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -59,6 +59,12 @@ else
 LDSCRIPT  = $(TOP_DIR)/memory.ld
 endif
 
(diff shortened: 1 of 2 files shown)


prusnak:
Thanks!

PR #186, opened 2017-07-01
Move bootloader and fastflash code to load_vector_table function

saleemrashid:
Removes duplicate code and uses a more idomatic method of loading the vector table.

### DIFF
--- a/util.h
+++ b/util.h
@@ -22,6 +22,9 @@
 
 #include <stdint.h>
 
+#include <libopencm3/cm3/scb.h>
+#include <libopencm3/cm3/vector.h>
+
 void delay(uint32_t wait);
 
 // converts uint32 to hexa (8 digits)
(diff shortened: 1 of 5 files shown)


util.h
```diff
@@ -38,4 +41,18 @@ void __attribute__((noreturn)) system_halt(void);
 // reset system
 void __attribute__((noreturn)) system_reset(void);
 
+static inline void __attribute__((noreturn)) load_vector_table(const vector_table_t *vector_table) {
+    // Relocate vector table
+    SCB_VTOR = (uint32_t) vector_table;
```
saleemrashid: Should this be masked with `0xFFFF`?

saleemrashid:
Should this be in `util.h`? It now requires `nvic.h` to be generated before compiling `libtrezor.a`

util.h
```diff
@@ -38,4 +41,18 @@ void __attribute__((noreturn)) system_halt(void);
...
+
+    // Set stack pointer
+    __asm__ volatile("msr msp, %0" :: "r" (vector_table->initial_sp_value));
+
+    // Jump to address
+    vector_table->reset();
+
+    // Prevent compiler from generating stack protector code (which causes CPU fault because the stack is moved)
```
saleemrashid: Should this be worded better?

util.h
prusnak: No

util.h
prusnak: I think it is fine.

prusnak:
Thanks!

PR #193, opened 2017-07-16
Add strwidth tool

saleemrashid:
Simple tool to check the width of strings (for development purposes)

```bash
gcc strwidth.c fonts.c -o strwidth -lreadline
```


prusnak:
Thanks!

PR #195, opened 2017-07-18
fix hangs due to using stale rng status

mruddy:
**High level info:**
This PR fixes hangs that are a result of how function `random32` mis-interprets (stale) RNG peripheral status.

Reference #153. This is the same fundamental problem. I closed #153  last time because it didn't seem to have momentum for getting merged. This time, I actually have experienced the hangs, can reproduce them, and have runtime gdb backtrace and register proof.

These device hangs can manifest themselves during signing messages, displaying number pads, resetting the device, switching between coins in the web wallet GUI, or simply booting up. I've experienced all of the aforementioned issues and tracked them to this single root cause.

**About the changes:**
1) first, this change does not set the interrupt enable bit in the rng control register. currently, the interrupt is not handled -- no need to set the bit.
2) second, DEBUG_RNG defaults to off since it's only a testing switch. Right now, the switch just makes it easier to quickly pull a bunch of entropy for quality testing.
3) **lastly, the bulk of the changes**: this change checks the SECS and CECS bits in the rng status register directly, rather than indirectly via SEIS and CEIS.
the way it appears to work is that, SECS and SEIS, and CECS and CEIS, match when an error is first detected.
but, SEIS and CEIS do not stay synched with the clock and fault detector as SECS and CECS do.
so, when an error occurs SEIS and CEIS go high and stay high because our code never resets them.
it appears to be the case that an error passes and SECS and CECS then go low again. that's why it's better to check them instead.
Excerpt from the reference manual section (20.4.2) on the RNG status register (RNG_SR):
```
Bit 2 SECS: Seed error current status
0: No faulty sequence has currently been detected. If the SEIS bit is set, this means that a faulty sequence was detected and the situation has been recovered.
1: One of the following faulty sequences has been detected:
   More than 64 consecutive bits at the same value (0 or 1)
   More than 32 consecutive alternances of 0 and 1 (0101010101...01)

Bit 1 CECS: Clock error current status
0: The RNG_CLK clock has been correctly detected. If the CEIS bit is set, this means that a clock error was detected and the situation has been recovered
1: The RNG_CLK was not correctly detected (f RNG_CLK < f HCLK /16).

Bit 0 DRDY: Data ready
0: The RNG_DR register is not yet valid, no random data is available
1: The RNG_DR register contains valid random data
```

**Debug environment and GDB info:**
The debugging environment for this was done according to the dev kit at: https://mcudev.github.io/trezor-dev-kit/index.html

Example GDB backtrace and register info captured during a "hang" while performing a device reset (r3 contains the RNG_SR value -- 0x41 -- meaning that "a faulty sequence was detected and the situation has been recovered" -- but the code hangs there anyways) The code gets stuck executing lines 0x8043264 - 0x804326a:
image

Here's another full backtrace taken while the hang was experienced during a message signing operation:
```
#0  random32 () at rng.c:30
#1  0x080213dc in generate_k_random (prime=0x80480c0 <secp256k1>, k=0x2001fd18) at ../vendor/trezor-crypto/ecdsa.c:201
#2  curve_to_jacobian (p=0x80481e4 <secp256k1+292>, jp=jp@entry=0x2001fcd0, prime=prime@entry=0x80480c0 <secp256k1>)
    at ../vendor/trezor-crypto/ecdsa.c:210
#3  0x08021f86 in scalar_multiply (curve=curve@entry=0x80480c0 <secp256k1>, k=k@entry=0x2001fd68, res=res@entry=0x2001fd8c)
    at ../vendor/trezor-crypto/ecdsa.c:607
#4  0x0802273c in ecdsa_get_public_key33 (curve=0x80480c0 <secp256k1>,
    priv_key=priv_key@entry=0x2000efa0 <node+40> "\224\065S\257\033\016\035T\205KH\221,f\356\366\227\063$\360}\026a]\342\324\316|\033\061\020m",
    pub_key=pub_key@entry=0x2000efc0 <node+72> "") at ../vendor/trezor-crypto/ecdsa.c:842
#5  0x080335c8 in hdnode_fill_public_key (node=0x2000ef78 <node>) at ../vendor/trezor-crypto/bip32.c:412
#6  hdnode_private_ckd (inout=inout@entry=0x2000ef78 <node>, i=0) at ../vendor/trezor-crypto/bip32.c:182
#7  0x08033882 in hdnode_private_ckd_cached (inout=inout@entry=0x2000ef78 <node>, i=i@entry=0x20002898 <msg_data+4>, i_count=i_count@entry=5,
    fingerprint=fingerprint@entry=0x0) at ../vendor/trezor-crypto/bip32.c:375
#8  0x08015944 in fsm_getDerivedNode (address_n_count=5, address_n=0x20002898 <msg_data+4>, curve=<optimized out>) at fsm.c:189
#9  fsm_msgSignMessage (msg=0x20002894 <msg_data>) at fsm.c:792
#10 0x08011b40 in MessageProcessFunc (ptr=<optimized out>, msg_id=<optimized out>, dir=<optimized out>, type=<optimized out>) at messages.c:64
#11 msg_process (msg_size=<optimized out>, msg_raw=0x20005894 <msg_in> "\b\254\200\200\200\b\b\200\200\200\200\b\b\200\200\200\200\b\b",
    fields=0x806ce50 <SignMessage_fields>, msg_id=38, type=110 'n') at messages.c:232
#12 msg_read_common (type=<optimized out>, buf=<optimized out>, len=<optimized out>) at messages.c:281
#13 0x0804456a in stm32fx07_poll (usbd_dev=0x20013f70 <usbd_dev>) at ../../usb/usb_fx07_common.c:348
#14 0x08010376 in usbPoll () at usb.c:414
#15 0x08012ea6 in main () at trezor.c:127
#16 0x080440f0 in reset_handler () at ../../cm3/vector.c:94
```

### DIFF
--- a/setup.c
+++ b/setup.c
@@ -42,7 +42,7 @@ void setup(void)
 
 	// enable RNG
 	rcc_periph_clock_enable(RCC_RNG);
-	RNG_CR |= RNG_CR_IE | RNG_CR_RNGEN;
+	RNG_CR |= RNG_CR_RNGEN;
 	// to be extra careful and heed the STM32F205xx Reference manual, Section 20.3.1
 	// we don't use the first random number generated after setting the RNGEN bit in setup
 	random32();
@@ -73,6 +73,9 @@ void setup(void)
 
 void setupApp(void)
 {
+	// for completeness, disable RNG peripheral interrupts for old bootloaders that had
+	// enabled them in RNG control register (the RNG interrupt was never enabled in the NVIC)
+	RNG_CR &= ~RNG_CR_IE;
 	// the static variables in random32 are separate between the bootloader and firmware.
 	// therefore, they need to be initialized here so that we can be sure to avoid dupes.
 	// this is to try to comply with STM32F205xx Reference manual - Section 20.3.1:
--- a/Makefile.include
+++ b/Makefile.include
@@ -65,6 +65,12 @@ else
 CFLAGS += -DMEMORY_PROTECT=1
 endif
 
+ifeq ($(DEBUG_RNG), 1)
+CFLAGS += -DDEBUG_RNG=1
+else
+CFLAGS += -DDEBUG_RNG=0
+endif
+
 LDFLAGS  += --static \
             -Wl,--start-group \
             -lc \
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -330,12 +330,14 @@ void fsm_msgWipeDevice(WipeDevice *msg)
 
 void fsm_msgGetEntropy(GetEntropy *msg)
 {
+#if !DEBUG_RNG
 	layoutDialogSwipe(&bmp_icon_question, _("Cancel"), _("Confirm"), NULL, _("Do you really want to"), _("send entropy?"), NULL, NULL, NULL, NULL);
 	if (!protectButton(ButtonRequestType_ButtonRequest_ProtectCall, false)) {
 		fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
 		layoutHome();
 		return;
 	}
+#endif
 	RESP_INIT(Entropy);
 	uint32_t len = msg->size;
 	if (len > 1024) {
--- a/rng.c
+++ b/rng.c
@@ -27,7 +27,7 @@ uint32_t random32(void)
 {
 	static uint32_t last = 0, new = 0;
 	while (new == last) {
-		if (((RNG_SR & (RNG_SR_SEIS | RNG_SR_CEIS)) == 0) && ((RNG_SR & RNG_SR_DRDY) > 0)) {
+		if ((RNG_SR & (RNG_SR_SECS | RNG_SR_CECS | RNG_SR_DRDY)) == RNG_SR_DRDY) {
 			new = RNG_DR;
 		}
 	}


prusnak:
LGTM, @jhoenicke @saleemrashid what do you think?

setup.c
```diff
@@ -42,7 +42,7 @@ void setup(void)
 
 	// enable RNG
 	rcc_periph_clock_enable(RCC_RNG);
-	RNG_CR |= RNG_CR_IE | RNG_CR_RNGEN;
+	RNG_CR |= RNG_CR_RNGEN;
```
prusnak: I think we might want to add `RNG_CR &= ~RNG_CR_IE;` to `setupApp()` function so we have the new behaviour also when using old bootloader and newer firmware, right?

jhoenicke:
@mruddy Thanks for investigating the issue.  The following is still not clear.

~~You enable the interrupt, but the interrupt handler is by default the blocking handler (see libopencm3: stm/f2/vector_nvic.c sets almost all isrs to the blocking handler).  This means an interrupt should freeze the mcu.  I wonder why you don't get freezes with the patch. ~~  Ooops, I read your patch in the wrong direction :)

The documentation says, the code should clear the `*EIS` bit after handling the problem.  Is it okay if we just ignore the bit and never clear it?  I think it should be okay.

setup.c
jhoenicke: Yes, I think so, too.

mruddy:
@jhoenicke The interrupt would be pending. I'm nearly certain that the RNG interrupt was never enabled, that would explain why no freezes all the time.

Yes, you're referring to section 20.3.2 ("What you should do is clear the SEIS bit, then clear and set the RNGEN bit to reinitialize and restart the RNG."). I wondered that same thing when reading it. But the doc seems to contradict itself in the description of the bits in section 20.4.2 "If the SEIS bit is set, this means that a faulty sequence was detected and the situation has been recovered." And also "If the CEIS bit is set, this means that a clock error was detected and the situation has been recovered". It seems to work just leaving the *EIS bits set and relying on the current status bits. But, yeah, the docs are not great in this area.

setup.c
mruddy: Good point. I'm sure that disabling the ENG_CR_IE does not affect whether SEIS will be set (or not). Not setting / unsetting this bit is just for completeness.

But it makes me wonder if we need to totally reset the RNG peripheral due to the order of operations in 20.3.1 where the interrupt is supposed to be enabled before enabling the peripheral, if you're going to use the RNG interrupt. I would guess it's probably safe to just unset RNG_CR_IE in setupApp though since we don't handle any pending RNG interrupts anyways.

mruddy: i went ahead and added the setupApp change (with comment). i have to run right now, but will research further and test later today. i think this is ok, though pending something unforeseen.

prusnak:
I just tested new code using suite in https://github.com/trezor/rng-test and so far it looks good.

mruddy:
@prusnak That's excellent news. Thanks for running those tests! I ran them back when working on #153 (which is nearly totally the same -- the same for the parts that matter for those tests) and also did not see any red flags. As far as I could tell, the quality of what was coming out of the RNG was just as good as before. I may run those tests again too myself later today when I get back to working on this.

jhoenicke:
Can this be the cause of https://www.reddit.com/r/TREZOR/comments/6nx5o4/installing_please_wait_take_forever/

The device checks a signature at that point, which calls random for the randomized ecc operations.

I  wonder why it doesn't happen for everyone.  It seems to happen consistently on some devices, while others are not affected.  Do some devices have a flaky random number generator?

prusnak:
Hmmm, I think I'll add RNG test to Bootloader SelfTest as I am working on that right now.

mruddy:
@jhoenicke It's plausible that this is related to that. I know for sure that generate_k_random in the backtrace above was called by a message signing operation and the comment on that code is "generate random K for signing/side-channel noise" which sounds very similar, if not the same, to what you are referencing.

This dev board of mine had been working without problems, then yesterday, it started hanging really often doing various things (while working in the web wallet). That's why I dug in this morning to see what was wrong. Every different hang came back to the random call.

I just got back to working on this and will do some more testing of this patch. I think it's good, but I just want to run some more tests. I have so far run a test with an old bootloader and this in a new firmware and did not have a problem, so that's a good sign.

prusnak:
Added RNG test to bootloader here: https://github.com/trezor/trezor-mcu/commit/bf374f17694707c2c754ff26b47c17c7480662bd

(use this via `trezorctl self_test` - this will be the "test suite" which will be run on TREZORs since the next batch)

mruddy:
@prusnak I'm not familiar with that, is that like a final step in device production that you guys run? I'll test it on my dev board and see if it fails without these changes. I assume the simple hope would be that the self-test does fail before, and then doesn't after this patch.

Also, I'm running the rng-test stuff. Waiting for dieharder right now. I remember that taking forever, so I'll probably kill it after one of them finishes.

For comparison, my dev board was taking fairly consistently about 21.5 seconds to generate 1MB of entropy:
`time ../python-trezor/tools/rng_entropy_collector.py stm32_rng_8.dat 1048576`

prusnak:
Yes, the SelfTest will be run on devices as a final step in device production.

It works (pre-patch) on boards I have at home, but I'll try it tomorrow on some returned devices I have at work.

mruddy:
I let dieharder run over two 1MB files. First file, everything passed except one test was weak. Second file, two tests said weak, but they were different than the first file. That looks pretty similar to the dieharder result files already in the repo, where most of the tests are pass and a couple are weak and change between input files.

Regarding the self-tester: I compiled the current master branch bootloader and firmware and put it on my dev board, then ran `trezorctl self_test` against it many times. I only got a hang a small number of times (on the order of maybe like 5% of the time). When it happened, the TESTING RNG... message just stayed on the display. Most of the time after reset or unplug/replug it passed again. It only failed twice in a row once. When it did hang, I verified with the debugger that the RNG_SR was 0x41 so this PR would let the test progress once the error clears and the status bits go low. I was hoping it would be reproducible more often. But, that's how it acted when I tested.

I'm going to see tomorrow morning if I can figure out a pattern for how to get the RNG into the SEIS state more predictably. Specifically, I want to see if there's something in the switch from bootloader to firmware that makes it happen more often, or not. If there is, then maybe we'd be better off re-initing the RNG when the firmware takes over. If I can't find anything, then this change is looking good to go.

prusnak:
I managed to find two boards which halted on almost every call to RNG before the patch and after the patch they are able to pass the RNG self-test multiple (10+ times).

Very good job @mruddy on analyzing and fixing the issue!

I will merge this issue and the next batch of bootloaders will include this change.

Thanks again!!!

mruddy:
Wow, great test result. Thanks to you guys for reviewing and testing this!

Just to follow-up on my remarks from last night on trying to figure out a pattern... I wrote a small bootloader that pretty much just read bytes from the RNG and incremented a counter to see how many iterations it took before the SEIS got set. My board doesn't seem to have a predictable pattern. Sometimes it failed quickly. Sometimes it took a minute or two and a couple hundred million iterations. Other times it took 9-10 minutes and overflowed the 32-bit counter. I also did some single stepping through the regular bootloader and firmware to see if there was something regularly causing the SEIS bit to get set, and I didn't see anything. So, I'm feeling good about these changes.

PR #196, opened 2017-07-24
NEM Support

saleemrashid:
All serialization and encryption code is in trezor/trezor-crypto#101, complete with unit tests. Compatible wallet software is ready.

Requires trezor/trezor-common#36


saleemrashid:
@prusnak I updated GCC to 4.9.3 for Travis CI, because the firmware was too large. Makes sense to use 4.9.3 because that's what the production firmware is built with, right?

saleemrashid:
Rebased and squashed, just needs trezor/trezor-common#86 to be merged now. As much as I would like to squash the `bn_format` refactor commit, it's too much work for too little gain IMO.

saleemrashid:
@prusnak Updated submodules and opened trezor/python-trezor#132 and trezor/connect#74

saleemrashid:
@prusnak Rebased (the Travis CI failure is due to #219)

prusnak:
Thank you Saleem for your hard work! Much appreciated!

PR #197, opened 2017-07-25
Clean-up of oled code.

jhoenicke:
Get rid of the macros to access pixels.  Use only the functions instead.
Simplified some of the drawing functions.

### DIFF
--- a/oled.c
+++ b/oled.c
@@ -63,18 +63,47 @@
  * display.
  */
 
-/* Macros to manipulate a single pixel in _oledbuffer:
- * OLED_BUFSET(X,Y) sets pixel X,Y (white)
- * OLED_BUFCLR(X,Y) clears pixel X,Y (black)
- * OLED_BUFTGL(X,Y) toggles pixel X,Y (inverts it)
- */
-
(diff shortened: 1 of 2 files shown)


saleemrashid:
I think there should be  `OLED_INDEX` and `OLED_MASK` macros to remove repetition of the same indexing and masking.

jhoenicke:
Forced push with the macros.  Code produced is the same.

saleemrashid:
I think you should remove `x` from `OLED_MASK(x, y)`

saleemrashid:
Also, would it be a good idea to make the arguments `unsigned` or even `uint8_t`?

jhoenicke:
OLED_MASK: it's intentional, otherwise you may accidently give x instead of y.
unsigned:  Most operations should cleanly handle out of screen access, e.g. draw centered string may run to the left and right outside the screen in exceptional cases (e.g. when signing MMMMMMMMMMMMMMM).

saleemrashid:
@jhoenicke Makes sense

prusnak:
Thanks!

PR #198, opened 2017-07-25
Bitcoin Cash

saleemrashid:
* Add `SIGHASH_FORKID` replay protection (requires `amount` in `TxInputType`)
* Add Bitcoin Cash support

trezor/trezor-common#37

### DIFF
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -396,6 +396,17 @@ bool compile_input_script_sig(TxInputType *tinput)
 			return false;
 		}
 	}
+	if (in_address_n_count >= 1) {
+		// check that input address didn't change
+		size_t count = tinput->address_n_count;
+		if (count != in_address_n_count) {
+			return false;
+		}
+		if (count >= 2
+			&& 0 != memcmp(in_address_n, tinput->address_n, (count - 2) * sizeof(uint32_t))) {
+			return false;
+		}
+	}
 	memcpy(&node, root, sizeof(HDNode));
 	if (hdnode_private_ckd_cached(&node, tinput->address_n, tinput->address_n_count, NULL) == 0) {
 		// Failed to derive private key
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -644,11 +644,6 @@ static bool signing_sign_input(void) {
 			signing_abort();
 			return false;
 		}
-		if (!input.has_amount) {
-			fsm_sendFailure(FailureType_Failure_DataError, _("SIGHASH_FORKID input without amount"));
-			signing_abort();
-			return false;
-		}
 		if (input.amount > to_spend) {
 			fsm_sendFailure(FailureType_Failure_DataError, _("Transaction has changed during signing"));
 			signing_abort();
@@ -825,7 +820,23 @@ void signing_txack(TransactionType *tx)
 					return;
 				}
 #endif
-				send_req_2_prev_meta();
+
+				if (coin->has_forkid) {
+					if (!tx->inputs[0].has_amount) {
+						fsm_sendFailure(FailureType_Failure_DataError, _("SIGHASH_FORKID input without amount"));
+						signing_abort();
+						return;
+					}
+					if (to_spend + tx->inputs[0].amount < to_spend) {
+						fsm_sendFailure(FailureType_Failure_DataError, _("Value overflow"));
+						signing_abort();
+						return;
+					}
+					to_spend += tx->inputs[0].amount;
+					phase1_request_next_input();
+				} else {
+					send_req_2_prev_meta();
+				}
 			} else if  (tx->inputs[0].script_type == InputScriptType_SPENDWITNESS
 						|| tx->inputs[0].script_type == InputScriptType_SPENDP2SHWITNESS) {
 				if (!coin->has_segwit || !coin->segwit) {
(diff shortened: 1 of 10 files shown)


prusnak:
@jhoenicke would you like to look at this?

jhoenicke:
Wow, just had a brief look over it.

Things to consider:
* The code enables replay protection for every bcc transaction (for Ethereum the signer can choose to use replay protection).  I think it makes sense to always use replay protection, so this is okay.
* Does it fix quadratic signing, does it check the amount?  For segwit inputs it doesn't ask for the previous transaction and it also doesn't request the whole tx for each input.

saleemrashid:
@jhoenicke It does check the amount, so I don't think we should check the previous transactions. Since you wrote the SegWit code, would it be possible for you to take over this PR?

saleemrashid:
@jhoenicke I will try and make it use a similar workflow to SegWit

saleemrashid:
@jhoenicke Could you take another look?

saleemrashid:
> The code enables replay protection for every bcc transaction (for Ethereum the signer can choose to use replay protection).

@jhoenicke https://github.com/Bitcoin-ABC/bitcoin-abc/commit/e49826c1fcc36e5ae26de0ad4d06e2063a759e73

jhoenicke:
I just pushed some changes to not rehash the current transaction for every input when forkid is set.  The patch got larger than I wanted, but I tried to remove some duplicated code.

@saleemrashid Can you write some unit tests for python_trezor similar to `test_msg_signtx.py`?  Would be nice if you do it with real transactions on bcc testnet.   Maybe also some tests to see if it detects changes in the amount on the second run.

I did not touch the merge problem in the protobuf directory.

saleemrashid:
@jhoenicke There isn't a Bitcoin Cash testnet yet, as far as I know. The only way I was testing this was simulating the UAHF with the Bitcoin-ABC client and validating the transactions.

jhoenicke:
Added some unit tests in jhoenicke/python-trezor@9917d9ebfc78593491c78bba0187c7330a15e6fb

prusnak:
Merged unit tests to python-trezor, Thanks!

Are there any pending changes you want to add to this branch? If not, I'll start the review and merge it afterwards.

jhoenicke:
I have no pending changes, so you can start the review :)

firmware/signing.c
```diff
@@ -1025,7 +1079,6 @@ void signing_txack(TransactionType *tx)
 				tx->inputs[0].script_sig.size = 0;
 			}
 			resp.serialized.serialized_tx.size = tx_serialize_input(&to, &tx->inputs[0], resp.serialized.serialized_tx.bytes);
-			update_ctr = 0;
```
jhoenicke: This is a minor  bug fix. The update_ctr counts when the gear should be redrawn.   Previously we had redrawn the gears here and reset update_ctr, after moving the redraw code around, this stuck here

firmware/signing.c
```diff
@@ -990,12 +1016,40 @@ void signing_txack(TransactionType *tx)
 			return;
 
 		case STAGE_REQUEST_SEGWIT_INPUT:
+			progress = 500 + ((signatures * progress_step) >> PROGRESS_PRECISION);
```
jhoenicke: I should remove this line. A similar line appears below in the right path.

We  don't want to update progress for segwit, where there is no progress for inputs (progress is counted in signatures and signatures are part  of the witnesses that come later).

firmware/signing.c
```diff
@@ -990,12 +1016,40 @@ void signing_txack(TransactionType *tx)
...
 			resp.has_serialized = true;
 			resp.serialized.has_signature_index = false;
 			resp.serialized.has_signature = false;
 			resp.serialized.has_serialized_tx = true;
-			if (tx->inputs[0].script_type == InputScriptType_SPENDP2SHWITNESS
-				&& !tx->inputs[0].has_multisig) {
+			if (tx->inputs[0].script_type == InputScriptType_SPENDADDRESS) {
```
jhoenicke: Here  is the new bcc case.  It's  input are treated like segwit inputs, so that they don't require to serialize the whole  transaction.

firmware/signing.c
```diff
@@ -777,10 +786,6 @@ void signing_txack(TransactionType *tx)
 			signing_check_input(&tx->inputs[0]);
 			if (tx->inputs[0].script_type == InputScriptType_SPENDMULTISIG
 				|| tx->inputs[0].script_type == InputScriptType_SPENDADDRESS) {
-				// remember the first non-segwit input -- this is the first input
```
jhoenicke: Don't make bcc inputs nonsegwit_inputs.  This way they use the segwit path that avoids serializing the whole transaction again.   This code moved below in the non-forkid path.

saleemrashid:
@jhoenicke Will we not support multisig with `SIGHASH_FORKID`?

jhoenicke:
I didn't test multisig yet.  Is there a reason why it wouldn't work?

saleemrashid:
@jhoenicke It's fine, it was some sort of weird issue with their full node implementation

jhoenicke:
We don't check for SPENDMULTISIG in `STAGE_REQUEST_SEGWIT_INPUT`.  I will change that, when I'm done with testing.

jhoenicke:
Added some bug fixes for multisig.

prusnak:
Thank you both! This is a great changeset!

PR #201, opened 2017-07-29
Do not wipe msg_resp in DebugLinkGetState

saleemrashid:
`protectButton` can call `fsm_msgDebugLinkGetState` in tiny mode, which uses `RESP_INIT`. This wipes the response to the original message.

Normally `protectButton` is called before writing to `resp` so this is not a visible issue. However, messages such as `GetAddress` call `protectButton` after.

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -1108,46 +1108,50 @@ void fsm_msgSetU2FCounter(SetU2FCounter *msg)
 void fsm_msgDebugLinkGetState(DebugLinkGetState *msg)
 {
 	(void)msg;
-	RESP_INIT(DebugLinkState);
 
-	resp->has_layout = true;
-	resp->layout.size = OLED_BUFSIZE;
-	memcpy(resp->layout.bytes, oledGetBuffer(), OLED_BUFSIZE);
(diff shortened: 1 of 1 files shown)


firmware/fsm.c
```diff
@@ -1108,46 +1108,51 @@ void fsm_msgSetU2FCounter(SetU2FCounter *msg)
...
 	(void)msg;
-	RESP_INIT(DebugLinkState);
 
-	resp->has_layout = true;
-	resp->layout.size = OLED_BUFSIZE;
-	memcpy(resp->layout.bytes, oledGetBuffer(), OLED_BUFSIZE);
+	// Do not use RESP_INIT because it clears msg_resp, but another message
+	// might be being handled
+	DebugLinkState resp;
+	_Static_assert(MSG_DEBUG_OUT_SIZE >= sizeof(resp), "DebugLinkGetState is too large");
```
jhoenicke: This line should be removed.  It is for checking that the msg_resp buffer is large enough, but it isn't used here.

saleemrashid:
@jhoenicke Fixed.

prusnak:
Thanks!

PR #202, opened 2017-07-31
Fix check for max try and add another check before checking the PIN.

jhoenicke:
If pin failure counter is too large wipe device instead of waiting and checking for PIN (in case the wipe after the last try failed.


prusnak:
Thanks!

PR #207, opened 2017-08-14
Assert that Features.coins max_count is large enough

saleemrashid:
Prevents silent crash in `fsm_msgGetFeatures` if `Features.coins` `max_count` is not large enough. Feel free to change the `_Static_assert` message.

/cc @jhoenicke


prusnak:
Good spot!

PR #209, opened 2017-08-14
Test matrix of DEBUG_LINK and FASTFLASH

saleemrashid:
This should catch inconsistencies with `Features.coins` or if there is a
build configuration that makes the firmware too large.

Also, add `MAKEFLAGS` and only build `STM32F2` support in `libopencm3` to
speed up build time


prusnak:
Thanks!

PR #212, opened 2017-08-20
Support multi-byte chain-id.

jhoenicke:
This pull request is not ready to be merged yet, but I want to put it here to not forget it.

The multi-byte chain-id is still untested.  When testing it, I found out, Musicoin doesn't support EIP-155 at all and will sign transactions that are not replay protected.  Also Musicoin doesn't have a BIP-44 id so it would be dangerous to use it with Trezor.  So I'd suggest to either remove the Musicoin case or convince them that they have to use EIP-155 and need a  BIP-44 prefix :)

BIP-44 prefixes for Expanse and UBIQ are not that well supported in myetherwallet.  I think it uses them by default, but there is no bullet point in the dialog to select them explicitly.  This is an issue of mew and its hardware wallet integration, though.

It looks like some geth local test configuration uses a multibyte chain id by default, maybe we can test the code with that.

### DIFF
--- a/firmware/ethereum.c
+++ b/firmware/ethereum.c
@@ -34,11 +34,14 @@
 #include "gettext.h"
 #include "ethereum_tokens.h"
 
+/* maximum supported chain id.  v must fit in an uint32_t. */
+#define MAX_CHAIN_ID 2147483630
+
 static bool ethereum_signing = false;
 static uint32_t data_total, data_left;
 static EthereumTxRequest msg_tx_request;
 static CONFIDENTIAL uint8_t privkey[32];
-static uint8_t chain_id;
+static uint32_t chain_id;
 struct SHA3_CTX keccak_ctx;
 
 static inline void hash_data(const uint8_t *buf, size_t size)
@@ -157,8 +160,17 @@ static void send_signature(void)
 
 	/* eip-155 replay protection */
 	if (chain_id != 0) {
+		uint8_t data[4];
+		data[0] = (chain_id >> 24) & 0xff;
+		data[1] = (chain_id >> 16) & 0xff;
+		data[2] = (chain_id >> 8) & 0xff;
+		data[3] = (chain_id) & 0xff;
+		int offset = 0;
+		while (!data[offset]) {
+			offset++;
+		}
 		/* hash v=chain_id, r=0, s=0 */
-		hash_rlp_field(&chain_id, 1);
+		hash_rlp_field(data+offset, 4-offset);
 		hash_rlp_length(0, 0);
 		hash_rlp_length(0, 0);
 	}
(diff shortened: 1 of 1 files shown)


firmware/ethereum.c
```diff
@@ -157,8 +160,17 @@ static void send_signature(void)
 
 	/* eip-155 replay protection */
 	if (chain_id != 0) {
+		uint8_t data[4];
+		data[0] = (chain_id >> 24) & 0xff;
+		data[1] = (chain_id >> 16) & 0xff;
+		data[2] = (chain_id >> 8) & 0xff;
+		data[3] = (chain_id) & 0xff;
+		int offset = 0;
+		while (!data[offset]) {
```
jhoenicke: This loop doesn't do bound checking.  It cannot leave the data array, since we know that chain_id is not zero, but I wonder if we should be more defensive here.

firmware/ethereum.c
```diff
@@ -225,6 +237,9 @@ static void ethereumFormatAmount(const bignum256 *amnt, const TokenType *token,
 			case  3: suffix = " tETH"; break;  // Ethereum Testnet: Ropsten
 			case  4: suffix = " tETH"; break;  // Ethereum Testnet: Rinkeby
 			case 42: suffix = " tETH"; break;  // Ethereum Testnet: Kovan
+			case  2: suffix = " EXP";  break;  // Expanse
+			case  8: suffix = " UBQ";  break;  // UBIQ
+			case 7762959: suffix = " MUSIC"; break; // Musicoin
```
jhoenicke: TODO: remove? (see pull request description)

runn1ng:
Slightly related: Musicoin support was added to MEW, but the PR was done wrong and shows users BTC XPUBs. 

Should we tell the users they cannot send their Musicoin and they sent it into a black hole? Or will it work with fixed paths (when they use BTC path for the account)

https://github.com/kvhnuke/etherwallet/pull/809

runn1ng:
(some users already tried to use it and sent their money there)

prusnak:
Frankly, it is irresponsible by MEW developers to add Musicoin in its current form.

jhoenicke:
> (some users already tried to use it and sent their money there, with BTC XPUBs)

I'm one of them... I used wireshark to find out which path mew was using.  Just 2.5 MUSIC, though, and it is on my test seed.

firmware/ethereum.c
prusnak: I think we should reuse `address_prefix_bytes_len` and `address_write_prefix_bytes` from https://github.com/trezor/trezor-crypto/blob/master/address.c here.

firmware/ethereum.c
prusnak: Yes, remove it for the time being.

firmware/ethereum.c
jhoenicke: Hmm, they really do the same, but I'm not sure if we should rely on that.  The idea of this code is, convert to big endian and strip leading zeros (the number encoding as specified in the Ethereum yellow paper).

Stcetjo:
Help! Sent coins to MEW + Trezor (when MEW first announced integration), now I can't find the address anymore using a "custom node". When I connect now, it doesn't show my original address. How do I look it back up to connect to it? My public address with musicoins: 0x99d915A7c195918C6F12d5E5be671F40C3D90824

PR #213, opened 2017-08-22
Improve old_storage_size logic

saleemrashid:
* Do not use hardcoded constants
* Do not include end padding
* More compact `if` statement (with `<=` instead of lots of `==`)

### DIFF
--- a/firmware/storage.c
+++ b/firmware/storage.c
@@ -148,16 +148,14 @@ bool storage_from_flash(void)
 	// copy storage
 	size_t old_storage_size = 0;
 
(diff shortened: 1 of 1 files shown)


prusnak:
Good idea. I have some rework of the storage in the queue, so I'll merge this later in a different fashion, but please keep it here for now.

jhoenicke:
Rebased this:

https://github.com/trezor/trezor-mcu/compare/master...jhoenicke:old-storage-size-improved

prusnak:
Will merge this via https://github.com/trezor/trezor-mcu/pull/270

PR #214, opened 2017-08-23
recovery: mark all related variables as confidential

romanz:
Following https://github.com/trezor/trezor-mcu/commit/98e617d8740b85ae01d7d6e0dd3f49e66057a210.

### DIFF
--- a/firmware/recovery.c
+++ b/firmware/recovery.c
@@ -35,59 +35,59 @@
 #include "recovery-table.h"
 
 /* number of words expected in the new seed */
-static uint32_t word_count;
+static CONFIDENTIAL uint32_t word_count;
 
 /* recovery mode:
  * 0: not recovering
  * 1: recover by scrambled plain text words
  * 2: recover by matrix entry
  */
-static int awaiting_word = 0;
+static CONFIDENTIAL int awaiting_word = 0;
 
 /* True if we should not write anything back to storage
  * (can be used for testing seed for correctness).
  */
-static bool dry_run;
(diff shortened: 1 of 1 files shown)


prusnak:
What exactly is the line of attack you are trying to prevent?

I don't think any of the variables used during recovery are confidential.

// cc @jhoenicke any thoughts on this?

romanz:
IIUC, after a dry-run recovery process is over, the `words`  variable (defined [here](https://github.com/trezor/trezor-mcu/pull/214/commits/9372d67804fa0bce1b15f49083cbebfed9cbc273?diff=split#diff-9c1db82f2b5e3cf122881188de852e6aR78)) contains the secret mnemonic. I thought that it also should be erased after a soft reset (by putting it into the `CONFIDENTIAL` linker section).

prusnak:
But they contain the mnemonic user just entered, no?

saleemrashid:
If the recovery is completed, these will be wiped from SRAM. If the recovery is not completed, we can assume that the user is present (e.g. it would far more likely they are physically attacked for the mnemonic words than the attacker extracts them from SRAM)

saleemrashid:
Also, variables like `awaiting_word`, `dry_run` and `enforce_wordlist` are not confidential in any way (and we should ensure that `.confidential` remains as small as possible)

romanz:
I agree with https://github.com/trezor/trezor-mcu/pull/214#issuecomment-325424089.
Closing this PR.

PR #217, opened 2017-08-30
Decouple flash-specific routines from storage

saleemrashid:
This makes it easier to use a non-flash storage.


prusnak:
NACK for now. I am working on a complete rewrite of storage, so let's keep this unmerged for now. (But also let's keep this open ...)

PR #218, opened 2017-08-31
memory_protect: add flash interface register check

mcudev:
I don't think this is urgent. This is more to be complete with the recommendations of some newly published research: https://www.usenix.org/system/files/conference/woot17/woot17-paper-obermaier.pdf

Section 3.2.4 on countermeasures: "This includes the “read protection level status” in the FLASH OBR register as well as the option bytes themselves."

FLASH_OPTCR is the f2 family equivalent flash interface register.

I was thinking of masking off the two low bits in my comparison, but then decided to go with the simple equality test because the option lock seems like it should always be set, and we should not be in the midst of an update to the option bytes, when making this check.

### DIFF
--- a/memory.c
+++ b/memory.c
@@ -30,8 +30,8 @@ void memory_protect(void)
 #if MEMORY_PROTECT
 	// Reference STM32F205 Flash programming manual revision 5 http://www.st.com/resource/en/programming_manual/cd00233952.pdf
 	// Section 2.6 Option bytes
-	//                     set RDP level 2                   WRP for sectors 0 and 1
-	if ((((*OPTION_BYTES_1) & 0xFFEC) == 0xCCEC) && (((*OPTION_BYTES_2) & 0xFFF) == 0xFFC)) {
+	//                     set RDP level 2                   WRP for sectors 0 and 1      flash option control register matches
+	if ((((*OPTION_BYTES_1) & 0xFFEC) == 0xCCEC) && (((*OPTION_BYTES_2) & 0xFFF) == 0xFFC) && (FLASH_OPTCR == 0x0FFCCCED)) {
 		return; // already set up correctly - bail out
 	}
 	flash_unlock_option_bytes();


prusnak:
Should not we do rather

```
(FLASH_OPTCR & 0x0FFFFFEC == 0x0FFCCCED)
```

instead of 

```
(FLASH_OPTCR == 0x0FFCCCED)
```

to test just the bits we care about?

saleemrashid:
@prusnak I think they covered that in the last part of the pull request comment

prusnak:
It's a little bit different as it's talking about masking two lower bits, not masking all reserved bits. But I agree the same reasoning could be applied to my question as well.

Let's ask @mruddy who did last change to this code, what he thinks about it.

saleemrashid:
@prusnak Sorry, I misread 0x0FFFFFEC as 0xFFFFFFFC :blush:

prusnak:
I'll merge this and do one extra commit.

prusnak:
Extra commit: https://github.com/trezor/trezor-mcu/commit/8ed9df66e178444ce6b7f75f285ad17b107ceefd

Please reopen when we need to change this. Thanks!

mruddy:
My input is that in 2.8.6 of PM0059 "STM32F205/215, STM32F207/217 Flash programming manual", the reserved bits 31-28 and 4 "must be kept cleared". With that in mind, I don't see a problem with using just the simple comparison.

Commit 0c30a193005f7121b6e38be208143f72abb8b26e looks like it has a bug. Masking off the low two bits will never allow the check to pass as it is: `(FLASH_OPTCR & 0x0FFFFFEC) == 0x0FFCCCED`

D means the low bit is set and will always fail. You could do `(FLASH_OPTCR & 0x0FFFFFEF) == 0x0FFCCCED` to ignore the reserved bits. That would match the style of the existing code.

prusnak:
Hm, yes, it's a bug, but should not we test rather using the following then?

```
(FLASH_OPTCR & 0x0FFFFFEC) == 0x0FFCCCEC
```

saleemrashid:
@prusnak That seems correct to me. From the programming manual:

> Bit 1 **OPTSTRT:** Option start
> This bit triggers a user option operation when set. It is set only by software and cleared when 
the BSY bit is cleared

So I don't think `FLASH_OPTCR_OPTSTRT` will be set there.

mruddy:
I think that goes back to what they said about the two low bits.

Verifying that the low two bits are 0 and 1 looks ok. It would be surprising to me if they equaled something else. The surprise would be that 1) an option operation would be in progress, and/or 2) the options are unlocked. If those bits are in a surprising state, then perhaps attempting the option rewrite is appropriate. Thus, the proposed simple equality check might be considered better.

For reference, here are the definitions:

```
Bit 1 OPTSTRT: Option start
This bit triggers a user option operation when set. It is set only by software and cleared when
the BSY bit is cleared.

Bit 0 OPTLOCK: Option lock
Write to 1 only. When this bit is set, it indicates that the FLASH_OPTCR register is locked.
This bit is cleared by hardware after detecting the unlock sequence.
In the event of an unsuccessful unlock operation, this bit remains set until the next reset.
```

saleemrashid:
@prusnak The equality test seems correct to me. From the programming manual:

> Bit 1 **OPTSTRT:** Option start
> This bit triggers a user option operation when set. It is set only by software and cleared when 
the BSY bit is cleared
> Bit 0 **OPTLOCK:** Option lock
> Write to 1 only. When this bit is set, it indicates that the FLASH_OPTCR register is locked. 
> This bit is cleared by hardware after detecting the unlock sequence.
> In the event of an unsuccessful unlock operation, this bit remains set until the next reset.

prusnak:
Ah, now I noticed that lower 2 bits in FLASH_OPTCR are not reserved like in OPTION_BYTES, but have different meaning and it really makes sense to what was proposed in the original commit.

Updated in https://github.com/trezor/trezor-mcu/commit/c796800b2b3c74f04bd936685a551c04fdd0b499

Thanks for the review!

mruddy:
You're welcome. Latest commit looks good.

mcudev:
Thanks to all for the code review!

PR #219, opened 2017-09-27
Generate the protobuf files at build time

jpz:
This is so as to improve the build reliability by ensuring protobuf files are generated at build time, to make life easier for future maintainers, and reduce ceremony of generating files and checking them in.

I've not been so opinionated as to remove the generated files from the repo as they get clobbered during the generation, but I do actually think it's a reasonable idea because it's very unambiguous that the developer should not hand-maintain them - if you comment on the PR that you would like this, I can add it to the change set.

Some notes: 

- I've pinned the nanopb submodule to the version that originally generated the files (0.2.9.3).
- In testing, I can confirm the files generated are identical down to the byte, to those presently in the repo.

I'd like to automate the generation of the coins files also in a future PR just as a tidy up, as I see some cut and manual generating and pasting of generated results occurring there also.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "libopencm3"]
 	path = vendor/libopencm3
 	url = https://github.com/libopencm3/libopencm3.git
+[submodule "vendor/nanopb"]
+	path = vendor/nanopb
+	url = https://github.com/nanopb/nanopb.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,5 @@ RUN apt-get update
 
 # install build tools and dependencies
 
-RUN apt-get install -y build-essential git python python-ecdsa gcc-arm-none-eabi
+RUN apt-get install -y build-essential git python python-ecdsa gcc-arm-none-eabi protobuf-compiler \
+	libprotobuf-dev python-protobuf python3-protobuf
--- a/build-firmware.sh
+++ b/build-firmware.sh
@@ -13,6 +13,8 @@ docker run -t -v $(pwd)/build:/build:z $IMAGE /bin/sh -c "\
 	git checkout $TAG && \
 	git submodule update --init && \
 	make -C vendor/libopencm3 && \
+	make -C vendor/nanopb/generator/proto && \
+	make -C firmware/protob && \
 	make && \
 	make -C firmware sign && \
 	cp firmware/trezor.bin /$BINFILE && \
--- a/firmware/protob/Makefile
+++ b/firmware/protob/Makefile
@@ -1,7 +1,7 @@
 all: messages.pb.c storage.pb.c types.pb.c messages_map.h
 
 %.pb.c: %.pb %.options
-	nanopb_generator.py $< -L '#include "%s"' -T
+	../../vendor/nanopb/generator/nanopb_generator.py $< -L '#include "%s"' -T
 
 %.pb: %.proto
 	protoc -I/usr/include -I. $< -o $@


prusnak:
Why do you install both python2 and python3 stuff (pip and protobuf)?

jpz:
Hi - thanks for the quick feedback.  What I found was:

`firmware/protob/messages_map.py' is a python3 program

`nanopb_generator.py` is a Python2 program.  It complains if you try to run it with python3.

I wanted to keep the surface area of changes down to what is easily reviewable, so didn't address this.  

Just looking at nanopb it looks like [later versions](https://github.com/nanopb/nanopb/issues/155) may support python3, which might get rid of the python 2 dependency.  

The syntax of the files changes quite a bit, and pb_* files might need to be updated if that change was made.  

That looks like a substantial change in its own right, worthy of it's own PR however.

jpz:
BTW, I installed pip to install protobuf - but it might be possible to use apt get python3-protobuf and skip pip entirely - it's just what I'm used to.

jpz:
I have removed pip/pip3 installs in preference to apt install of same components.

saleemrashid:
> nanopb_generator.py is a Python2 program. It complains if you try to run it with python3.

While 0.2.9.3 is not, the latest version of nanopb is definitely compatible with Python 3.

Unfortunately, we cannot safely upgrade nanopb because the storage structure would change. For example, nanopb would change the representation for `optional bytes homescreen`  in `Storage` from using `size_t` to, with `-DPB_FIELD_16BIT`, a `uint16_t`.

saleemrashid:
```diff
--- nanopb-0.2.9.3.txt
+++ nanopb-master.txt
@@ -1,51 +1,47 @@
 struct _Storage {
 	uint32_t                   version;              /*     0     4 */
 	_Bool                      has_node;             /*     4     1 */
 
 	/* XXX 3 bytes hole, try to pack */
 
-	HDNodeType                 node;                 /*     8   132 */
-	/* --- cacheline 2 boundary (128 bytes) was 12 bytes ago --- */
-	_Bool                      has_mnemonic;         /*   140     1 */
-	char                       mnemonic[241];        /*   141   241 */
-	/* --- cacheline 5 boundary (320 bytes) was 62 bytes ago --- */
-	_Bool                      has_passphrase_protection; /*   382     1 */
-	_Bool                      passphrase_protection; /*   383     1 */
-	/* --- cacheline 6 boundary (384 bytes) --- */
-	_Bool                      has_pin_failed_attempts; /*   384     1 */
+	HDNodeType                 node;                 /*     8   120 */
+	/* --- cacheline 2 boundary (128 bytes) --- */
+	_Bool                      has_mnemonic;         /*   128     1 */
+	char                       mnemonic[241];        /*   129   241 */
+	/* --- cacheline 5 boundary (320 bytes) was 50 bytes ago --- */
+	_Bool                      has_passphrase_protection; /*   370     1 */
+	_Bool                      passphrase_protection; /*   371     1 */
+	_Bool                      has_pin_failed_attempts; /*   372     1 */
 
 	/* XXX 3 bytes hole, try to pack */
 
-	uint32_t                   pin_failed_attempts;  /*   388     4 */
-	_Bool                      has_pin;              /*   392     1 */
-	char                       pin[10];              /*   393    10 */
-	_Bool                      has_language;         /*   403     1 */
-	char                       language[17];         /*   404    17 */
-	_Bool                      has_label;            /*   421     1 */
-	char                       label[33];            /*   422    33 */
-	/* --- cacheline 7 boundary (448 bytes) was 7 bytes ago --- */
-	_Bool                      has_imported;         /*   455     1 */
-	_Bool                      imported;             /*   456     1 */
-	_Bool                      has_homescreen;       /*   457     1 */
-
-	/* XXX 2 bytes hole, try to pack */
-
-	Storage_homescreen_t       homescreen;           /*   460  1028 */
-	/* --- cacheline 23 boundary (1472 bytes) was 16 bytes ago --- */
-	_Bool                      has_u2f_counter;      /*  1488     1 */
+	uint32_t                   pin_failed_attempts;  /*   376     4 */
+	_Bool                      has_pin;              /*   380     1 */
+	char                       pin[10];              /*   381    10 */
+	/* --- cacheline 6 boundary (384 bytes) was 7 bytes ago --- */
+	_Bool                      has_language;         /*   391     1 */
+	char                       language[17];         /*   392    17 */
+	_Bool                      has_label;            /*   409     1 */
+	char                       label[33];            /*   410    33 */
+	_Bool                      has_imported;         /*   443     1 */
+	_Bool                      imported;             /*   444     1 */
+	_Bool                      has_homescreen;       /*   445     1 */
+	Storage_homescreen_t       homescreen;           /*   446  1026 */
+	/* --- cacheline 23 boundary (1472 bytes) --- */
+	_Bool                      has_u2f_counter;      /*  1472     1 */
 
 	/* XXX 3 bytes hole, try to pack */
 
-	uint32_t                   u2f_counter;          /*  1492     4 */
-	_Bool                      has_needs_backup;     /*  1496     1 */
-	_Bool                      needs_backup;         /*  1497     1 */
-	_Bool                      has_flags;            /*  1498     1 */
+	uint32_t                   u2f_counter;          /*  1476     4 */
+	_Bool                      has_needs_backup;     /*  1480     1 */
+	_Bool                      needs_backup;         /*  1481     1 */
+	_Bool                      has_flags;            /*  1482     1 */
 
 	/* XXX 1 byte hole, try to pack */
 
-	uint32_t                   flags;                /*  1500     4 */
+	uint32_t                   flags;                /*  1484     4 */
 
-	/* size: 1504, cachelines: 24, members: 25 */
-	/* sum members: 1492, holes: 5, sum holes: 12 */
-	/* last cacheline: 32 bytes */
+	/* size: 1488, cachelines: 24, members: 25 */
+	/* sum members: 1478, holes: 4, sum holes: 10 */
+	/* last cacheline: 16 bytes */
 };
```

jpz:
@saleemrashid - thanks for the analysis.  Yes, I thought there was too many changes in the generated code by nanopb when I ran against the latest version, and that diff proves the intuition was right to pin it to the old version.

When I was working on the repo manually just a day to two ago, the first instinct was to grab the latest nanopb, and rerun the generator to see what happens.  If I had made a change to .proto, I could have easily generated the protobuf with the wrong version, and submitted a poor PR.

By pinning nanopb to 0.2.9.3 as I've done in this PR however, it should make it very straightforward for any future altcoin contributors to just "get it right" and avoid this type of error, as the repo will provide the nanopb install and version with a GitHub submodule, and you can read the Makefile to see what is going on.

If you're happy with the contribution, the next question for you as the maintainers is whether the generated files should be removed from the repo in this PR.  On one hand it's nice to have a reference copy of them, on the other, it's reduces ambiguity from future altcoin contributions, and will avoid the possibility of inconsistency.

A final suggestion/alternative is that I just leave the working Makefile generation as a task which is run only manually (rather than build time in the Dockerfile), if you prefer that approach, which accomplishes 50% or so of the benefits.

saleemrashid:
@jpz We should be able to switch to the latest version of nanopb once the storage code is refactored (it uses the generated structure as the on-flash format).

Also, could you update `.travis.yml` with the new `make` commands (should just be `make -C vendor/nanopb/generator/proto` and `make -C firmware/protob`)?

prusnak:
@saleemrashid I just fixed the `.travis.yml`, sorry for not noticing earlier :-/

There is one outstanding issue, but out of scope of this pull request.

PR #220, opened 2017-09-28
Generate coins.h and coins.c source code at build time

jpz:
This generates the coins.h and coins.c files from coins.json source at build time, to help ensure consistency.

Previously it was reliant upon the user to run coins-gen.py manually and hand-edit the coins.h and coins.c files.

I'm checking the two new generated files in here `coins_array.h` and `coins_count.h`, however that is just for illustration - the build process clobbers them consistently.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -139,7 +139,7 @@ $(NAME).elf: $(OBJS) $(LDSCRIPT) $(TOOLCHAIN_DIR)/lib/libopencm3_stm32f2.a $(TOP
 %.small.o: %.c Makefile
 	$(CC) $(CFLAGS) -MMD -o $@ -c $<
 
-clean:
+clean::
 	rm -f $(OBJS)
 	rm -f *.a
 	rm -f *.bin
--- a/firmware/.gitignore
+++ b/firmware/.gitignore
@@ -1,0 +1,2 @@
+coins_array.h
+coins_count.h
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -95,3 +95,18 @@ bootloader.o: ../fastflash/bootloader.bin
 		--redefine-sym _binary_$(shell echo -n "$<" | tr -c "[:alnum:]" "_")_size=__bootloader_size__ \
 		--rename-section .data=.rodata \
 		$< $@
+
+# ensure header files are generated prior to compiling sources
+coins.c crypto.c fsm.c transaction.c: coins_count.h
+coins.c: coins_array.h
+
+#################
+# Code Generation
+coins_count.h: coins-gen.py coins.json
+	./$< count > $@
+
+coins_array.h: coins-gen.py coins.json
+	./$< array > $@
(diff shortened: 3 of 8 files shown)


prusnak:
Please, don't include generated files in commit.

Also is `clean::` in Makefile.include a typo?

jpz:
Oops - Travis failed - I believe this is due trying to build master of `git clone http://github.com/trezor-mcu` which is run inside docker.

This was another aspect I considered it might be worth to clean up in the build, which would be to have Docker mount the locally cloned repo as a file system, and build that.  At present the build is slightly like a race condition in that you can't be absolutely sure it will work as committing to master changes the test results itself.

jpz:
Cool - regarding `clean::` - this was something I had to look up in Makefile documentation.

To have `clean:` work to delete the count_* files in the firmware directory, make has to process two clean directives.  The single colon one will give a syntax error if it sees a redefinition.  The double-colon version acts in an additive way.  Thus you get the rm of the count files, and the rms specified in Makefile.include run as well, when cleaning inside the firmware directory.

Will delete the generated files.

jpz:
actually - still need to check the Travis log a bit more - Travis is not using docker.

jpz:
Ok, the build is all fixed.  I've rebased to master, so it fixes the build issue your recent check in had also on master, which was actually related. 

I've broken it the PR into two discrete changesets - the automatic generation of the files, and the fixing of Travis build process.

I wasn't able to get `make -C firmware` to build the protob directory, the `make -C firmware/protob` needs to be staged first.  As the Makefile in the subdirectory generates C files, it's not trivial to get make to re-parse the dependency list after the addition of new files.

jpz:
I removed the fixes for .travis.yml and python, as I see you've solved that on the main branch.

Note I changed the make parameter to -j1 - at -j2 in the .travis.yml, as otherwise it can start compiling .c files into .o files prior to finishing the code generation, due to parallelising the work.

I also wanted to discuss the generation of the Ethereum files.  I did not add this to this particular PR, for two reasons.  1) I think the coins generation work stands on its own in any case, and 2) the ethereum generation code pulls stuff from an non-version-pinned external source - so automating the script as it stands could lead to non-reproduceable builds until a more locked down strategy might be decided.

prusnak:
> Note I changed the make parameter to -j1 - at -j2 in the .travis.yml, as otherwise it can start compiling .c files into .o files prior to finishing the code generation, due to parallelising the work.


This should not be a problem if dependency is properly expressed in the makefile (i.e. that  `firmware/coins.c` depends on `firmware/coins_array.h` and `firmware/coins.h` depends on `firmware/coins_count.h`). I would rather have this fixed than disabling the parallel build.

Let's not do the same for Ethereum. The source is non-authoritative and not under our control.

jpz:
Cool this is done.

prusnak:
Looks good! Thanks!

PR #221, opened 2017-09-29
Dockerfile: Use multi-line instructions

saleemrashid:
Minimize the number of layers and call `apt-get update` properly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ script:
   - make
   - make -C bootloader
   - make -C fastflash
+  - sed -i 's/, deprecated = true//' firmware/protob/messages.proto # protobuf 2.5 compatibility :-/
   - make -C firmware/protob
   - make -C firmware
   - make -C demo
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,8 @@
 
 FROM debian:9
 
-# update repositories
-
-RUN apt-get update
-
 # install build tools and dependencies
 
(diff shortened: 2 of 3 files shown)


prusnak:
I don't like this. I prefer to have packages, that are installed to allow some logical part, to be grouped together. That way when you change one line you don't have to reinstall everything (just the stuff that follows).

The other issue has already been fixed.

saleemrashid:
> when you change one line you don't have to reinstall everything (just the stuff that follows)

That's the issue. You need to have the `apt-get update` in the same command as `apt-get install` to prevent caching problems. See the [best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run).

Relevant part:

> Always combine `RUN apt-get update` with `apt-get install` in the same `RUN` statement.
> 
> Using `apt-get update` alone in a `RUN` statement causes caching issues and subsequent `apt-get install` instructions fail.

Also, it is recommended to minimize build layers. If you want to change packages for development, you could just add a new `RUN` command temporarily.

prusnak:
Message fields are sth different than Enum fields.

prusnak:
Fixed via https://github.com/trezor/trezor-mcu/commit/4b8ac90d105ee8ffeb65d7d5d30ccf6e98891f7c

Now to merge your stuff

prusnak:
Merged via https://github.com/trezor/trezor-mcu/commit/4147a914ac2030ef357c30bb9d7e728bed2f378c

PR #223, opened 2017-10-07
NEM Importance Transfer Transaction Support

saleemrashid:
Requires trezor/trezor-common#46 and trezor/trezor-crypto#114


prusnak:
Thanks!

PR #224, opened 2017-10-09
Generate nem_mosaics.[ch] at build time

saleemrashid:
See #220


saleemrashid:
Well, _that_ was fun. But using Protocol Buffers 3.4.0 makes sense and adds mere seconds to the build time.

prusnak:
Thanks!

PR #225, opened 2017-10-10
Makefile.include: Generate dependency files before build

saleemrashid:
Generate dependency files before the build (so we do not have to specify generated header file dependencies manually)

 * `-MP` is used to generate empty rules for header files (so we don't have to remove stale `.d` files if a header file is deleted)
 * `-MG` is used to make GCC ignore missing header files (so we can create dependencies on generated header files such as `nem_mosaics.h`)

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -101,14 +101,6 @@ bootloader.o: ../fastflash/bootloader.bin
 		--rename-section .data=.rodata \
 		$< $@
 
-# ensure header files are generated prior to compiling sources
-coins.c crypto.c fsm.c transaction.c: coins_count.h
-coins.c: coins_array.h
-
-nem2.c layout2.c fsm.c: nem_mosaics.h
-
-#################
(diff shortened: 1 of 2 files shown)


firmware/Makefile
```diff
@@ -101,14 +101,6 @@ bootloader.o: ../fastflash/bootloader.bin
...
 		$< $@
 
-# ensure header files are generated prior to compiling sources
-coins.c crypto.c fsm.c transaction.c: coins_count.h
-coins.c: coins_array.h
-
-nem2.c layout2.c fsm.c: nem_mosaics.h
-
-#################
-# Code Generation
```
saleemrashid: Removing this comment was personal preference (it seemed a bit ugly to me) but feel free to keep it

prusnak:
Thanks!

jpz:
Great stuff - I those rules by hand there, and couldn't figure out a way of doing it more tidily.

PR #227, opened 2017-10-12
Fix checking change address

### LINKED ISSUE #226: mixing segwit/non-segwit inputs is broken
... because of how we hangle bip44/bip49 paths in extract_input_bip32_path/check_change_bip32_path of [signing.c](https://github.com/trezor/trezor-mcu/blob/master/firmware/signing.c)

jhoenicke:
There was a signed/unsigned problem: size_t is unsigned, but we use
-1 to indicate mismatch.  The problem was that when checking the input
address path, it still did this unintentionally when a mismatch was
detected, forbidding to sign with mismatched inputs, even when there
is no change address.

We now use 1 for mismatch.  Also we don't allow change address anymore
if the inputs have a path of length 1.  This simplifies the code a bit.

This fixes #226

### DIFF
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -68,6 +68,12 @@ static uint8_t multisig_fp[32];
 static uint32_t in_address_n[8];
 static size_t in_address_n_count;
 
+/* marker for in_address_n_count to indicate a mismatch in bip32 paths in 
+   input */
+#define BIP32_NOCHANGEALLOWED 1
+/* maximum allowed change index */
+#define MAX_BIP32_LAST_ELEMENT 1000000
+
 enum {
 	SIGHASH_ALL = 1,
 	SIGHASH_FORKID = 0x40,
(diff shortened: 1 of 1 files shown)


saleemrashid:
> There was a signed/unsigned problem: size_t is unsigned, but we use -1 to indicate mismatch.

In future, we should use `ssize_t` for this.

firmware/signing.c
```diff
@@ -328,61 +334,45 @@ void phase2_request_next_input(void)
 
 void extract_input_bip32_path(const TxInputType *tinput)
 {
-	if (in_address_n_count == (size_t) -1) {
+	if (in_address_n_count == BIP32_NOCHANGEALLOWED) {
 		return;
 	}
 	size_t count = tinput->address_n_count;
-	if (count < 1) {
+	if (count < 2) {
```
saleemrashid: What about `if (count <= BIP32_NOCHANGEALLOWED)`?

jhoenicke: The logic here is that bip32 paths should follow the ..../0/x or .../1/x rule for wallets, so there should be at least two levels.  The change address should have the last two levels replaced with 1/x.

I want to keep the option to change BIP32_NOCHANGEALLOWED to something else (like -1), without having to change this line.

saleemrashid: What about adding `#define BIP32_MINLENGTH 2` or similar?

jhoenicke: Hmm, maybe something like `BIP32_WALLETDEPTH`, to indicate that this are the number of levels that change inside a single wallet. if we want to go this way, we should also have `#define BIP32_CHANGECHAIN 1`.  It is still not fully configurable, e.g. in `check_change_bip32_path` we need to check each of these two levels for validity.

prusnak:
Thanks!

PR #234, opened 2017-10-29
Enable SIGHASH_FORKID for SIGVERSION_BASE and SegWit

saleemrashid:
This is not tested against Bitcoin Gold but it has not broken Bitcoin or Bitcoin Cash. All hardcoded instances of `SIGHASH_ALL` have been removed.

/cc @jhoenicke

### DIFF
--- a/firmware/signing.c
+++ b/firmware/signing.c
@@ -733,6 +733,8 @@ static bool signing_sign_segwit_input(TxInputType *txinput) {
 		resp.has_serialized = true;
 		if (!signing_sign_hash(txinput, node.private_key, node.public_key, hash))
 			return false;
+
+		uint8_t sighash = signing_hash_type() & 0xff;
 		if (txinput->has_multisig) {
 			uint32_t r = 1; // skip number of items (filled in later)
 			resp.serialized.serialized_tx.bytes[r] = 0; r++;
@@ -742,7 +744,7 @@ static bool signing_sign_segwit_input(TxInputType *txinput) {
 					continue;
 				}
 				nwitnesses++;
-				txinput->multisig.signatures[i].bytes[txinput->multisig.signatures[i].size] = 1;
+				txinput->multisig.signatures[i].bytes[txinput->multisig.signatures[i].size] = sighash;
 				r += tx_serialize_script(txinput->multisig.signatures[i].size + 1, txinput->multisig.signatures[i].bytes, resp.serialized.serialized_tx.bytes + r);
 			}
 			uint32_t script_len = compile_script_multisig(&txinput->multisig, 0);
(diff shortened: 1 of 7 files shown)


jhoenicke:
It looks good, but I didn't test it.

When you test it, can  you add unit tests to python-trezor?  Similar to `test_msg_signtx_bch.py`, but with both segwit and non-segwit transactions.  Of course this requires to add a coin definition.

saleemrashid:
@jhoenicke I will try and add tests that work against Bitcoin Gold. The coin definition is in trezor/trezor-common#51

saleemrashid:
@runn1ng Bitcoin Gold with the replay protection PR referenced by trezor/trezor-common#51

jhoenicke:
I noticed that BTG currently uses for non-segwit transactions old-style signatures (non-bip143) with forkid added.  I added a comment to their repository, but If they don't change this, this is something we need to handle.

saleemrashid:
Thanks for catching that @jhoenicke. However, if they don't change it, this should be easy to handle. Instead of using `forkid` in this pull request, I could add `forkid_segwit`.

saleemrashid:
@jhoenicke I can't find your comment on their repository, could you link it here? Anyway, this is the relevant line of code https://github.com/BTCGPU/BTCGPU/pull/83/files#diff-be2905e2f5218ecdbe4e55637dac75f3R1274

jhoenicke:
I may have added it to one of the commits.  I tried to add it again (a few lines above the position you suggested).  Hope that this works better.

saleemrashid:
Here's the approach I am going to take now:

* Refactor out all the hardcoded `nHashType` values (I want to do this whether or not we need this for BTG)
* Add a `forkid_bip143` or similar option to the coin definition or similar to decide whether to use BIP143 signature hashes for non-SegWit transactions (I think `force_bip143` might be a better name though as there's no real dependency on `forkid`)

saleemrashid:
This is **entirely untested** but, if it works, we can handle Bitcoin Gold now and if they follow @jhoenicke's suggestion to use BIP143 for non-SegWit transactions.

Requires trezor/trezor-common#51 for `force_bip143`

jhoenicke:
I looked at the code yesterday and it took me ages to find where we add the hashtype to the hash. I think we should change this and do this in signing.c before computing the digest.  I'm preparing a patch

saleemrashid:
@jhoenicke I fixed that in the first commit in this pull request (I think)

jhoenicke:
This patch is what I meant.  It is just ugly that tx_hash_output adds the hash_type, that is not even part of the transaction.

firmware/signing.c
```diff
@@ -658,6 +668,7 @@ static bool signing_sign_hash(TxInputType *txinput, const uint8_t* private_key,
 	}
 	resp.serialized.signature.size = ecdsa_sig_to_der(sig, resp.serialized.signature.bytes);
 
+	uint8_t sighash = signing_hash_type() & 0x1f;
```
jhoenicke: I think this should be 0xff.  Not that we want to support SIG_ANYONECANPAY soon, but still logically it's the last byte, right?

jhoenicke:
Adding `signing_hash_type()` was a good idea BTW, the code looks much cleaner now.

jhoenicke:
Bch and btc transactions are now tested and work. I guess it doesn't make sense to test btg transactions before they at least merge the pull request.

saleemrashid:
Rebased, squashed the `signing_hash_type() & 0xff` fix and dropped my `add_hash_type` fix.

saleemrashid:
Untested rebase

jhoenicke:
Tested on gold testnet and fixed a bug.
Didn't test multisig.

saleemrashid:
@jhoenicke I modified your last commit to make it consistent with `signing_sign_hash`

prusnak:
Thank you!

PR #235, opened 2017-10-30
New CoinInfo separated from protobuf structures

jhoenicke:
Having CoinType using the protobuf structures has several disadvantages.
- We always need to change trezor-common if we need a new field (like bech32 prefix)
- Every time Trezor initializes it sends all this information out and nobody cares.
- The protobuf structures add storage overhead due to their fixed size.

I also removed some of the `has_` fields:
- `has_segwit` and `segwit` are now merged.
- `has_coin_shortcut` can be replaced by test for NULL if necessary.

### DIFF
--- a/firmware/coins-gen.py
+++ b/firmware/coins-gen.py
@@ -13,34 +13,18 @@ if len(sys.argv) != 2 or sys.argv[1] not in ("count", "array"):
 
 def get_fields(coin):
     return [
-        'true' if coin['coin_name'] is not None else 'false',
         '"%s"' % coin['coin_name'] if coin['coin_name'] is not None else 'NULL',
-
-        'true' if coin['coin_shortcut'] is not None else 'false',
         '" %s"' % coin['coin_shortcut'] if coin['coin_shortcut'] is not None else 'NULL',
-
-        'true' if coin['address_type'] is not None else 'false',
-        '%d' % coin['address_type'] if coin['address_type'] is not None else '0',
-
-        'true' if coin['maxfee_kb'] is not None else 'false',
         '%d' % coin['maxfee_kb'] if coin['maxfee_kb'] is not None else '0',
-
-        'true' if coin['address_type_p2sh'] is not None else 'false',
-        '%d' % coin['address_type_p2sh'] if coin['address_type_p2sh'] is not None else '0',
-
-        'true' if coin['signed_message_header'] is not None else 'false',
         '"\\x%02x" "%s"' % (len(coin['signed_message_header']), coin['signed_message_header'].replace('\n', '\\n')) if coin['signed_message_header'] is not None else 'NULL',
-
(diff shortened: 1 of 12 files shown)


jhoenicke:
Things to discuss:

- Is it a good idea to remove the has_address_type...  fields?  Should we use a different value than 0 to check for presence? (which doesn't work for address_type,  since bitcoin uses 0).
- Which fields of CoinInfo should `fsm_getFeatures` send to the PC?

firmware/coins-gen.py
```diff
@@ -15,14 +15,16 @@ def get_fields(coin):
     return [
```
saleemrashid: Can we refactor this to use named struct fields?

firmware/fsm.c
```diff
@@ -208,6 +208,8 @@ void fsm_msgInitialize(Initialize *msg)
 	fsm_msgGetFeatures(0);
 }
 
+_Static_assert(pb_arraysize(Features, coins) >= COINS_COUNT, "Features.coins max_count not large enough");
```
saleemrashid: Maybe we should move this to the start of the file?

saleemrashid:
Related: instead of `memcpy` and `FT_STATIC`, we could use `FT_POINTER` for `Features.coins`. Less controversial than this, at least temporarily.

jhoenicke:
I force-pushed to combine the first two commits.  I added another commit moving the static assert and adding segwit to getFeatures.

For the other fields, I cannot imagine anybody is using them.

firmware/fsm.c
```diff
@@ -56,6 +56,8 @@
 #include "rfc6979.h"
 #include "gettext.h"
 
+_Static_assert(pb_arraysize(Features, coins) >= COINS_COUNT, "Features.coins max_count not large enough");
```
saleemrashid: Sorry to be irritating about this, but I just realised you could put the `_Static_assert` inside `fsm_msgGetFeatures` which might look neater.

saleemrashid:
> For the other fields, I cannot imagine anybody is using them.

I don't think we can safely assume that. For example, the last open source version of TREZOR Wallet was using `address_type` (https://github.com/trezor/webwallet/blob/1ddbe6559bf60271fb42fe9e4a21e89053a4f12b/app/scripts/services/TrezorAccount.js#L25).

saleemrashid:
> we could use `FT_POINTER` for `Features.coins`

I was wrong about this, because `FT_POINTER` does not allow you to use `const` pointers. We could type-pun (as we know we never call `pb_decode` on it) but it's safer to use `memcpy`.

jhoenicke:
@saleemrashid Do you think it is less controversial if I make it fully backwards compatible, i.e., fill all the fields we supported so far?  Though I really would like to not send the "Bitcoin Signed Message" header for every supported currency.

prusnak:
Thanks!

PR #236, opened 2017-10-31
Enable Segwit Bech32 addresses

jhoenicke:
This is built on top of the coininfo patches, see #235.  Thus these are included here.
The new patches start at 971e236.

You need to set bech32_prefix in coins.json to "bc" resp. "tb"  for bitcoin/testnet to test this.

And you need to update trezor-crypto to my bech32 branch (this is why checks fail).

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -694,7 +694,7 @@ void fsm_msgGetAddress(GetAddress *msg)
 		}
 		bool qrcode = false;
 		for (;;) {
-			layoutAddress(address, desc, qrcode);
+			layoutAddress(address, desc, qrcode, msg->script_type == InputScriptType_SPENDWITNESS);
 			if (protectButton(ButtonRequestType_ButtonRequest_Address, false)) {
 				break;
 			}
@@ -732,7 +732,7 @@ void fsm_msgEthereumGetAddress(EthereumGetAddress *msg)
 
 		bool qrcode = false;
 		for (;;) {
-			layoutAddress(address, desc, qrcode);
+			layoutAddress(address, desc, qrcode, false);
 			if (protectButton(ButtonRequestType_ButtonRequest_Address, false)) {
 				break;
 			}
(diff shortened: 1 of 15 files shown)


jhoenicke:
The output confirmation dialog when signing still doesn't handle the longer addresses properly.
I'd like to use the "to" line to get three lines for the address.

prusnak:
> I'd like to use the "to" line to get three lines for the address.

Yeah, let's do it. Some Zcash addresses also don't fit (because they use 2 bytes for address version).

jhoenicke:
> Yeah, let's do it. Some Zcash addresses also don't fit (because they use 2 bytes for address version).

Done in db2ad06.  Very long addresses like Segwit's  p2wsh addresses (62 characters) still don't fit, but only 5 characters or so are missing, so it shouldn't affect security.

Saleem suggested to make the dialog scroll down each time the user presses confirm, but if we go this way, we should make it a general feature instead of hacking it into ConfirmOutput.

prusnak:
Looks nice. Are these final changes? If yes I'll review and merge soon.

firmware/fsm.c
```diff
@@ -1135,7 +1154,7 @@ void fsm_msgNEMGetAddress(NEMGetAddress *msg)
 
 		bool qrcode = false;
 		for (;;) {
-			layoutAddress(resp->address, desc, qrcode);
+			layoutAddress(resp->address, desc, qrcode, false);
```
jhoenicke: @saleemrashid If NEM only uses upper case/numbers/hyphen in address, we can change this to true to get QR codes with better error correction level.

saleemrashid: ACK (it's Base32)

jhoenicke:
@prusnak Yes they are final.  You may want to review and merge #235  first, this should remove the patches from this pull request :)

prusnak:
merged in 600c61b3dcbfeb10e90e3f776b0b0bf0c3f6a7f3

firmware/fsm.c
prusnak: Applied in 892bb8501a1282535a9c721d0a3187f51aa5585b

PR #239, opened 2017-11-01
Disable swipe with DEBUG_LINK

saleemrashid:
This greatly reduces the amount of time that device tests take. With SegWit native device test, it reduced the time by 40%.

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -38,10 +38,19 @@
 
 void *layoutLast = layoutHome;
 
+void layoutSwipe(void) {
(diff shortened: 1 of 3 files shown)


firmware/layout2.c
```diff
@@ -38,10 +38,19 @@
...
 
+void layoutSwipe(void) {
+#if DEBUG_LINK
+	oledClear();
+#else
+	oledSwipeLeft();
+#endif
+}
+
+
```
saleemrashid: Sorry, I'll have to fix the extra newline later

PR #240, opened 2017-11-03
Support for generic ECDSA signing

zulucrypto:
I've implemented generic ECDSA signing because I'd like to add support for signing Stellar transactions. This PR adds a generic method that I'll use to build out additional features for Stellar in the future.

`fsm_msgSignIdentity` came close to what I wanted, but it didn't support an arbitrary bip32 path or provide a way to confirm the hash of the data being signed.

This also relies on a PR submitted to trezor-common: https://github.com/trezor/trezor-common/pull/53

### DIFF
--- a/firmware/crypto.c
+++ b/firmware/crypto.c
@@ -110,6 +110,12 @@ int gpgMessageSign(HDNode *node, const uint8_t *message, size_t message_len, uin
 	}
 }
 
+int ecdsaMessageSign(HDNode *node, const uint8_t *message, size_t message_len, uint8_t *signature)
+{
+	signature[0] = 0; // prefix: pad with zero, so all signatures are 65 bytes
+	return hdnode_sign(node, message, message_len, signature + 1, NULL, NULL);
+}
+
 int cryptoMessageSign(const CoinType *coin, HDNode *node, InputScriptType script_type, const uint8_t *message, size_t message_len, uint8_t *signature)
 {
 	SHA256_CTX ctx;
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -927,6 +927,62 @@ void fsm_msgSignIdentity(SignIdentity *msg)
 	layoutHome();
 }
 
+void fsm_msgSignEcdsa(SignEcdsa *msg)
+{
+	RESP_INIT(SignedEcdsa);
+
+	CHECK_INITIALIZED
+
+	layoutSignEcdsa(msg);
+	if (!protectButton(ButtonRequestType_ButtonRequest_ProtectCall, false)) {
+		fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
+		layoutHome();
+		return;
+	}
+
+	CHECK_PIN
+
+	HDNode *node = fsm_getDerivedNode(msg->ecdsa_curve_name, msg->address_n, msg->address_n_count);
+	if (!node) {
+		fsm_sendFailure(FailureType_Failure_DataError, _("Unable to get node, invalid curve or address"));
+		layoutHome();
+		return;
+	}
+
+	int result = 0;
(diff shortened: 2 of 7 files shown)


firmware/crypto.c
```diff
@@ -110,6 +110,12 @@ int gpgMessageSign(HDNode *node, const uint8_t *message, size_t message_len, uin
 	}
 }
 
+int ecdsaMessageSign(HDNode *node, const uint8_t *message, size_t message_len, uint8_t *signature)
```
zulucrypto: This is the same as `sshMessageSign`, but I used this name for clarity. Let me know if you'd prefer I remove `sshMessageSign` and replaced it with this function.

firmware/layout2.c
```diff
@@ -419,6 +420,24 @@ void layoutSignIdentity(const IdentityType *identity, const char *challenge)
 		NULL);
 }
 
+void layoutSignEcdsa(const SignEcdsa *signEcdsa)
```
zulucrypto: This displays the SHA256 of the data being signed in order for the user to verify they're signing what they expect.

saleemrashid:
> This PR adds a generic method that I'll use to build out additional features for Stellar in the future.

You shouldn't need this pull request to implement Stellar support in the firmware though.

firmware/protob/messages.options
```diff
@@ -137,6 +137,15 @@ SignedIdentity.address			max_size:60
 SignedIdentity.public_key		max_size:33
 SignedIdentity.signature		max_size:65
 
+SignEcdsa.address_n             max_count:8
+SignEcdsa.blob		            max_size:1024
+SignEcdsa.challenge_visual		max_size:256
```
saleemrashid: What is this field used for?

firmware/fsm.c
```diff
@@ -927,6 +927,62 @@ void fsm_msgSignIdentity(SignIdentity *msg)
...
+		return;
+	}
+
+	// Assemble public key
+	hdnode_fill_public_key(node);
+	if (strcmp(msg->ecdsa_curve_name, SECP256K1_NAME) != 0) {
+		resp->has_address = false;
+	} else {
+		resp->has_address = true;
+		hdnode_get_address(node, 0x00, resp->address, sizeof(resp->address)); // hardcoded Bitcoin address type
```
saleemrashid: What is the use of this?

zulucrypto:
> NACK. This subverts the security of the TREZOR by allowing to sign, for example, Bitcoin transactions, messages, etc. without the user's knowledge.

@saleemrashid Would you be willing to accept a similar PR if it was specific to signing Stellar transactions?

In detail:

1. Remove `ecdsaMessageSign` and replace with `stellarSign`
2. Force BIP32 path to start with m/44'/148'/ (Stellar Lumen BIP-44 coin type from https://github.com/satoshilabs/slips/blob/master/slip-0044.md)
3. Remove `layoutSignEcdsa` and replace with `layoutSignStellar` with a Stellar-specific UI

Thank you for your quick review and comments!

saleemrashid:
> Would you be willing to accept a similar PR

I don't work for TREZOR, I just contribute code! :smile:

It's unlikely a PR like you described would be accepted though because it doesn't provide the security benefits of a hardware wallet (e.g. displaying transaction details). However, if you implement full Stellar support similarly to Ethereum or NEM support, it's very likely @prusnak will accept it. But you may want to discuss with him.

prusnak:
What is the format of the transaction being signed? (i.e. what data do you send in `blob`?) That would be the first step in understanding what you are trying to achieve and create a StellarSignTx message.

prusnak:
Also, let's move discussion here: https://github.com/trezor/trezor-common/pull/53 where we iterate the StellarSignTx message and I'll close this request for the time being.

Issue #232, opened 2017-10-26
Tx fee estimation can be wrong with op_return


runn1ng:
When transaction has OP_RETURN outputs, the calculation of fee can be wrong

https://github.com/trezor/trezor-mcu/blob/d006ef6bf7edfa2334fa4d03f004aa09d072526a/firmware/signing.c#L594

OP_RETURN output scripts have variable length (from I think 2 up to 41 bytes), this calculation asumes fixed output length.

I am not sure if this can be used for some attack or not

runn1ng:
Perhaps somehow chained with #231 ?

runn1ng:
OP_RETURN is not displayed, so attacker can modify the transaction to include thousands of OP_RETURNs and the user won't detect it, and will force it to pay larger fees?

saleemrashid:
@runn1ng You can have no more than one `OP_RETURN` output per transaction and each `OP_RETURN` output is limited to 80 bytes. Therefore, the most that an attacker could add to the transaction would be 90 bytes (8 bytes for amount, 1 byte for script length, 1 byte for `OP_RETURN`, 80 bytes of data)

saleemrashid:
That would mean that the estimated transaction size would be off by no more than 55 bytes (which seems acceptable to me). Also, we're underestimating the size rather than overestimating, so we would calculate the fee rate as higher rather than lower and be more likely to warn the user.

runn1ng:
Oh, there is just one op-return per tx, I forgot that

runn1ng:
I am closing this, since this seems fine

runn1ng:
> You can have no more than one OP_RETURN output per transaction and each OP_RETURN output is limited to 80 bytes

Hm, that only applies to what is standard, but Trezor still lets you make the transaction, right. It won't relay, but it stil valid transaction, right?

saleemrashid:
@runn1ng While that's true, the below still applies so I don't think this is an issue

> Also, we're underestimating the size rather than overestimating, so we would calculate the fee rate as higher rather than lower and be more likely to warn the user.

runn1ng:
OP_RETURN can also use empty data, which would cause overestimating.

jhoenicke:
Is someone working on this? If not I give it a shot.  It should not be too difficult to get an almost exact estimate (except for the signature length).

prusnak:
@jhoenicke Yes, please give it a shot.

prusnak:
Fixed via #242

PR #242, opened 2017-11-08
Byte-precise size estimate for fees

jhoenicke:
Fixes issue #232.
**Note**: This patch displays the size estimate on the TREZOR display for debugging purposes.  I intend to remove this later.   It is ready for review/testing,  but I want to remove this before merging.

It assumes largest possible signature size for all inputs.  For segwit multisig it can show a .25 bytes larger value per input due to difference between segwit encoding (varint) vs. non-segwit encoding (op_push) of the multisig script. The segwit discount is included in the calculations.

### DIFF
--- a/firmware/transaction.c
+++ b/firmware/transaction.c
@@ -35,19 +35,41 @@
 
 #define SEGWIT_VERSION_0 0
 
+/* transaction input size (without script): 32 prevhash, 4 idx, 4 sequence */
+#define TXSIZE_INPUT 40
+/* transaction output size (without script): 8 amount */
+#define TXSIZE_OUTPUT 8
+/* size of a pubkey */
+#define TXSIZE_PUBKEY 33
+/* size of a DER signature (3 type bytes, 3 len bytes, 33 R, 32 S, 1 sighash */
+#define TXSIZE_SIGNATURE 72
+/* size of a multiscript without pubkey (1 M, 1 N, 1 checksig) */
+#define TXSIZE_MULTISIGSCRIPT 3
+/* size of a p2wpkh script (1 version, 1 push, 20 hash) */
+#define TXSIZE_WITNESSPKHASH 22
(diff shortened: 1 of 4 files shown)


jhoenicke:
Tested with some of the unit tests and manual inspection of the transaction and computation of its size.

firmware/transaction.c
```diff
@@ -35,19 +35,41 @@
...
+#define op_push_size(len) ((len) < 0x4c ? 1 : (len) < 0x100 ? 2 : \
+						   (len) < 0x10000 ? 3 : 5)
+
 uint32_t op_push(uint32_t i, uint8_t *out) {
 	if (i < 0x4C) {
 		out[0] = i & 0xFF;
 		return 1;
 	}
-	if (i < 0xFF) {
+	if (i < 0x100) {
```
jhoenicke: This actually fixes a bug; Trezor would have produced non bip-62 conformant push for exactly 255 bytes. 
OTOH push is only used for OP_RETURN or multisig,  OP_RETURN is  limited to 80 bytes and multisig cannot be exactly 255 bytes (it is always 3 + 34*pubkey_count)

firmware/transaction.c
```diff
@@ -35,19 +35,41 @@
...
 		out[0] = 0x4C;
 		out[1] = i & 0xFF;
 		return 2;
 	}
-	if (i < 0xFFFF) {
+	if (i < 0x10000) {
```
jhoenicke: In principle we could remove the case >= 0x10000.  Pushes beyond 520 bytes are not allowed in bitcoin.

firmware/transaction.c
prusnak: ACK, good find.

firmware/transaction.c
prusnak: Let's leave it there.

jhoenicke:
TODOs removed and force pushed.

prusnak:
Thanks!

PR #245, opened 2017-11-17
Confirm lock time on device

### LINKED ISSUE #233: locktime is not displayed on device
When transaction has locktime, it's not displayed on device.

Adding locktime to non-locktime transaction is not a problem IMO (worse that could happen is that user will not send his coins).

However, attacker can *strip* locktime (or lower it) when user wants a large one.

saleemrashid:
Asks user to confirm lock time as block height (`block #1`) or date (`01 Jan 1970 00:00 UTC`).

Does not display seconds and only supports UTC.

Fixes #233

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -20,6 +20,7 @@
 #include <stdint.h>
 #include <string.h>
 #include <ctype.h>
+#include <time.h>
 
 #include "layout2.h"
 #include "storage.h"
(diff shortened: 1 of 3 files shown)


prusnak:
The change is nice, but as per the last comment in #233 we decided not to show lock time on the device as this breaks user experience on Electrum for example (which always uses last known height as locktime). There is no known attack except striping the lock time info, which is not a big issue.

saleemrashid:
@prusnak Argh, didn't see that. Sorry :blush:

prusnak:
Let's discuss if you feel there is a need for this because it prevents a serious problem.

prusnak:
I am also sorry for not merging this otherwise perfectly legit commit :-/

Issue #51, opened 2015-11-13
Use non-proportional font for addresses


prusnak:
... and maybe another parts of info. See also https://github.com/trezor/trezor-mcu/issues/33

prusnak:
Implemented in https://github.com/trezor/trezor-mcu/pull/295

PR #246, opened 2017-11-18
Ask user to confirm passphrase

### LINKED ISSUE #62: show passphrase on Trezor
Instead of typing the passphrase twice each time, user should enter the passphrase only once.

User should be able to show the passphrase on TREZOR instead on PC

saleemrashid:
Asks the user to confirm the passphrase before using it.

Perhaps we should not allow the user to cancel the passphrase as the UX with TREZOR Wallet is terrible. Rather than asking again for the passphrase, TREZOR Wallet will keep sending the same passphrase unless the device is reattached or the page reloaded.

Fixes #62

### DIFF
--- a/firmware/protect.c
+++ b/firmware/protect.c
@@ -35,6 +35,8 @@
 
 bool protectAbortedByInitialize = false;
 
+static uint8_t CONFIDENTIAL passphraseHash[SHA256_DIGEST_LENGTH];
+
 bool protectButton(ButtonRequestType type, bool confirm_only)
 {
 	ButtonRequest resp;
@@ -237,9 +239,6 @@ bool protectChangePin(void)
 	return result;
 }
 
-
-static uint8_t passphraseHash[32];
-
 bool protectPassphrase(void)
 {
 	if (!storage_hasPassphraseProtection() || session_isPassphraseCached()) {
@@ -262,10 +261,10 @@ bool protectPassphrase(void)
 
 			size_t length = strlen(ppa->passphrase);
 			if (length > 0) {
-				uint8_t hash[32];
+				uint8_t hash[SHA256_DIGEST_LENGTH];
 				sha256_Raw((const uint8_t *)(ppa->passphrase), length, hash);
 
-				bool sameAsLast = memcmp(hash, passphraseHash, 32) == 0;
+				bool sameAsLast = memcmp(hash, passphraseHash, SHA256_DIGEST_LENGTH) == 0;
 
 				if (!sameAsLast) {
 					layoutPassphrase(ppa->passphrase, length);
--- a/firmware/protect.c
+++ b/firmware/protect.c
@@ -255,6 +255,16 @@ bool protectPassphrase(void)
 		if (msg_tiny_id == MessageType_MessageType_PassphraseAck) {
 			msg_tiny_id = 0xFFFF;
 			PassphraseAck *ppa = (PassphraseAck *)msg_tiny;
+
+			size_t length = strlen(ppa->passphrase);
+			if (length > 0) {
+				layoutPassphrase(ppa->passphrase, length);
+				if (!protectButton(ButtonRequestType_ButtonRequest_ProtectCall, true)) {
+					result = false;
+					break;
+				}
+			}
+
 			session_cachePassphrase(ppa->passphrase);
 			result = true;
 			break;
--- a/firmware/protect.c
+++ b/firmware/protect.c
@@ -29,6 +29,7 @@
 #include "util.h"
 #include "debug.h"
 #include "gettext.h"
+#include "sha2.h"
 
 #define MAX_WRONG_PINS 15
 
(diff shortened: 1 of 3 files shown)


prusnak:
Will discuss with the team on Monday. Thanks!

Two discuss points:

a) do we want to show the dialog if the provided passphrase is empty? (e.g., password manager always sends empty passphrase)
b) should the protectButton be cancellable?

jhoenicke:
Can we add a field, e.g., `confirmOnDevice` to `PassphraseAck` message?  This way the behaviour can be triggered by  PC (if it supports it).  I'm not sure if an extra byte or two fits into the "tiny" packet, though.

saleemrashid:
@jhoenicke NACK. We should keep it consistent to reduce confusion.

I think the least controversial choice would be to go with "no" for both of @prusnak's points.

Regarding tiny packets, we could rewrite them to use the same code as normal packets, essentially removing their limitations.

jhoenicke:
> We should keep it consistent to reduce the confusion.

But the wallet could still keep it consistent.  The wallet could see that the Trezor supports this, only asks for the passphrase once and always enables confirmation on device.  And an old wallet that doesn't support the feature would just work the same way it always worked, because it doesn't set the flag. This way you avoid the problem with cancelling and also the old wallet doesn't get an unexpected ButtonRequest.

Or are you worried that a malicious wallet may disable this feature to hide that it plays tricks with the passphrase?

saleemrashid:
> only asks for the passphrase once

I disagree with this sentiment of #62. I think we should continue to ask the user to confirm on the computer.

> the old wallet doesn't get an unexpected ButtonRequest.

Software should be designed to always expect `ButtonRequest` so I don't think this is an issue

> Or are you worried that a malicious wallet may disable this feature to hide that it plays tricks with the passphrase?

Yes, this is part of the issue.

saleemrashid:
Force-pushed least controversial option https://github.com/trezor/trezor-mcu/pull/246#issuecomment-345548394

runn1ng:
I have not tested this, but. There might be a problem with current implementation in trezor web wallet

Right now, on start of every action, we call "init", which erases the passphrase, and then send the passphrase, which is remembered in browser in plaintext. If we added this (if it works how I think), the user will need to confirm passphrase before every action.

It might be usable, I am not sure. I would need to try it. :)

Also, does this correctly deals with whitespace? (for example spaces after the passphrase, or 2 spaces instead of 1, or tabs/space in passphrase)

karel-3d:
I have tried it on our web wallet, it works fine. We should probably remove "double" passphrase with this

saleemrashid:
@karel-3d We could store a hash of the last non-empty passphrase in memory and compare the new passphrase to that.

karel-3d:
We could, but now that I am thinking about this again and actually trying it, it's not that terrible, and we would introduce possible bugs with that.

saleemrashid:
@karel-3d I don't think we can really introduce bugs with that. Also, "not that terrible" != "great" :smile:

karel-3d:
There are still possible attacks on this, with whitespace. You can add spaces after the passphrase and it still looks like the original passphrase. (I am not sure how limited this attack is, I am not sure how long the passphrase can be.)

saleemrashid:
@karel-3d Well, a power user will know that there are at most 16 characters per line so the attack is somewhat limited :smile:

More seriously, this is a good start but you raise a good point and it may be worth revisiting #51

karel-3d:
I think that we should not wait with #51 for this; even with whitespace bugs, this is better than the current state

karel-3d:
I made a pull request to this pull request :) here

https://github.com/saleemrashid/trezor-mcu/pull/1

firmware/protect.c
```diff
@@ -262,10 +261,10 @@ bool protectPassphrase(void)
 
 			size_t length = strlen(ppa->passphrase);
 			if (length > 0) {
-				uint8_t hash[32];
+				uint8_t hash[SHA256_DIGEST_LENGTH];
 				sha256_Raw((const uint8_t *)(ppa->passphrase), length, hash);
 
-				bool sameAsLast = memcmp(hash, passphraseHash, 32) == 0;
+				bool sameAsLast = memcmp(hash, passphraseHash, SHA256_DIGEST_LENGTH) == 0;
```
saleemrashid: Non-constant time check here shouldn't be a problem because you would need to do a pre-image attack to exploit it. And the user would notice all the passphrase confirmations!

saleemrashid:
Rebased on `master` and merged previous confirmed hash checks. Thanks @karel-3d!

karel-3d:
With the device state logic here - https://github.com/trezor/trezor-mcu/commit/7fa8ae136fc45ddfae94ff54be8fd763dc6b159a - saving and checking of passphrase hash is no longer needed, since that is what device state does. (At least that is my understanding of the feature.)

prusnak:
We'll be redesigning passphrase in the new monorepo: https://github.com/trezor/trezor-firmware/issues/1

PR #255, opened 2017-12-02
Add IOTA support

bartslinger:
Requires https://github.com/trezor/trezor-common/pull/61

IOTA uses completely different cryptography compared to other coins. I still derive the 'iota seed' from a HDNode, using path `m/IOTA'` with curve ED25519. This curve has no parameters, so effectively it just applies hmac_sha512. I use the output private key (32 bytes) and chain code (32 bytes) to form a 64-byte deterministic seed that I in turn use to generate a 81-tryte IOTA seed. With this method, I hoped to accomplish two things:
1. Deterministic IOTA addresses from a 24 words mnemonic
2. No security breach for other coins on the trezor if the IOTA seed gets compromised.

Generating public addresses on trezor takes around 12 seconds per address. I added an address counter in storage which keeps track of used addresses. The alternative would be to generate addresses until a balance is found. For each previous transaction, this would add 12 seconds to the login process, which I consider undesirable.

I'm not sure if it is correct to add 8 bytes for a uint32_t and a bool in firmware/storage.c

For signing transactions, the computer already has to know the send address, change address and corersponding address indices. This should be implemented in the computer side wallet.

The IOTA transaction should also include a relatively accurate timestamp. Depending on how long the user takes to verify the transaction by pushing buttons, the timestamp might get old. I implemented the following process to get a relatively accurate timestamp:
1. The user sends a transaction request to the trezor, including the current timestamp.
2. The user has to verify all the transaction details on the trezor, including that timestamp.
3. The approval message is sent to the computer.
4. The computer responds with the newest timestamp.
5. The transaction signing is initiated using the user verfied data together with the newest timestamp. This timestamp should be within 5 minutes from the initial one.
6. Once sigining is complete, the signatures are sent to the computer.

This is a video of the workflow https://www.youtube.com/watch?v=HK8jIogYgcI

For the computer side, I basically only modified https://github.com/iota-trezor/python-trezor to allow command line interaction. The goal is to integrate trezor support in upcoming IOTA wallets.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "vendor/nanopb"]
 	path = vendor/nanopb
 	url = https://github.com/nanopb/nanopb.git
+[submodule "vendor/iota"]
+	path = vendor/iota
+	url = https://github.com/iota-trezor/iota.git
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -26,6 +26,7 @@ OBJS += ethereum.o
 OBJS += ethereum_tokens.o
 OBJS += nem2.o
 OBJS += nem_mosaics.o
+OBJS += iota.o
 
 OBJS += debug.o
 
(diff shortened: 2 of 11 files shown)


bartslinger:
I made some changes that reflect the feedback in https://github.com/trezor/trezor-common/pull/61

The signing process is much simpler now:
1. The user sends a transaction request to the trezor, including a timestamp which can be a little bit into the future.
2. The user has to verify all the transaction details on the trezor, including that timestamp.
3. The transaction is signed and its signatures are sent to the computer.

I completely removed the option to show the IOTA seed on the display.

The transaction request does not include the option anymore to use custom address indices. The trezor decides on that. The user can still (re)set the address counter, after a confirmation on the trezor. The address counter can be requested by the computer and will be shown on the screen. This allows for recovering. The computer also receives the current address counter.

bartslinger:
It has been a while. I would like to have a green light on this before proceeding with the pc side.

bartslinger:
I've lost my interest in this project. Doesn't seem to go anywhere.

PR #261, opened 2017-12-09
Generic hash support for Bitcoin transactions

saleemrashid:
Rewrite Bitcoin transaction signing to support other hash algorithms. This is useful for coins like Decred, which uses BLAKE256 instead of SHA256.

The only remaining SHA256 specific code in the transaction signing is the `ecdsa_get_pubkeyhash` and `ecdsa_get_address_*` calls.

All device tests pass.

/cc @peterzen


prusnak:
> All device tests pass.

Can confirm this. LGTM! Thanks!

PR #262, opened 2017-12-09
Generic curve and hash support for Bitcoin coins

saleemrashid:
See trezor/trezor-mcu#261 for rationale

At the moment I have hardcoded `'secp256k1'` in `coins-gen.py` but I will change that in a future pull request. For Decred, for example, we can have:

```c
const curve_info secp256k1_decred_info = {
    /* This is not standard though, will need to check with the Decred team */
    .bip32_name = "Decred seed",
    .params = &secp256k1,
    .hasher_type = HASHER_BLAKE,
};
```

Where `HASHER_BLAKE` is a `HasherType` for BLAKE256.

All device tests pass.

Requires trezor/trezor-crypto#118


peterzen:
Awesome, thank you @saleemrashid !

saleemrashid:
@peterzen Could you confirm that it's alright to use `Decred seed` instead of `Bitcoin seed`? My rationale is that Decred uses BLAKE256 and Bitcoin uses SHA256 so we shouldn't allow a compromise of one hash function to affect the other coin.

peterzen:
I think these questions are best directed to @davecgh

davecgh:
I'm not sure about the first question since it looks like it only applies to the Trezor code which I'm not really familiar enough with to know all of the consequences.  It looks like it's just a name though, so I don't think there would be any problem.

In terms of differences in the transactions the main differences are:

- The previous outpoint of transaction inputs includes an additional `int8` field which refers to which transaction tree it goes into (regular or stake)
- The transaction inputs include additional witness fields for the input amount, prevout block height (`uint32`), and prevout block index (position within the block, `uint32`) that were populated by the miner.
- The transaction outputs include an addition `uint16` field for the script version
- The transaction includes an additional `uint32` field for expiration

saleemrashid:
> I'm not sure about the first question since it looks like it only applies to the Trezor code which I'm not really familiar enough with to know all of the consequences. It looks like it's just a name though, so I don't think there would be any problem.

Basically, importing a BIP39 seed from the TREZOR would not work with other Decred software. However, you could import an XPRV or trivially modify the Decred software to solve this.

> In terms of differences in the transactions the main differences are:

Thanks for the information. Could you be more specific about the witness serialization?

peterzen:
`dcrwallet` uses a PGP word list, not BIP39 mnemonics so the alternative seed should not be a problem.

davecgh:
The current Decred wallets don't use BIP39 rather they use the PGP wordlist, so they won't be able to import BIP39 seeds from Trezor without modifications anyways.  Since seeds are all just a stream of bytes under the hood anyways, I don't see any issue with that.  It'll just be a matter of decoding the BIP39 encoding to bytes.

I'll answer the serialization question in another response.

saleemrashid:
@davecgh I think I understand the witness data serialization now but now I'm wondering about the signature hash. Shall I only implement `SIGHASH_ALL_VALUE` (that would make signing faster because the hardware wallet doesn't have to ask for previous transactions).

saleemrashid:
@davecgh To be honest, it would just be generally helpful if you explain the transaction serialization and signature hash in more depth :smile:

davecgh:
Here is the struct for a transaction input:

```Go
type TxIn struct {
	// Non-witness
	PreviousOutPoint OutPoint
	Sequence         uint32

	// Witness
	ValueIn         int64
	BlockHeight     uint32
	BlockIndex      uint32
	SignatureScript []byte
}
```

The signature hash is a single blake256 hash over the serialization of `signature_hash_type || prefix_hash || witness_hash`.

The prefix hash is a single blake256 hash of the serialization of `version|(serialize_type<<16) || num_txin || all_txin_prefixes || num_txout || all_txouts || lock_time || expiration`

The txin prefixes are the serialization of `prevout_hash || prevout_index || prevout_tree || sequence` and the txout serialization consists of `value || script_version || public_key_script_len || public_key_script`.

The witness hash is a single blake256 hash of the serialization of  either `version|(serialize_type<<16) || num_txin || all_txin_witness_signing_fields` or `version|(serialize_type<<16) || num_txin ||  all_txin_witness_value_signing_fields` depending on the signature hash type (second only applies to `SigHashAllValue`).

The witness signing fields referenced above consist of `public_key_script_len || public_key_script` and the witness value signing fields are `amount || public_key_script_len || public_key_script`.

As should be obvious from the breakdown, this allows the signature to either commit to the input amounts of not depending in the signature hash type.  This is useful for cold/offline wallet devices such as trezor since it means they don't require the full tx being spent since they can safely learn the input amount from an untrusted source (for example the `ValueIn` field of the transaction input) and any invalid amounts would be detected as an invalid signature by the network.

saleemrashid:
@davecgh What hash would an outpoint be referencing? The prefix hash?

davecgh:
Correct.  It's the prefix hash so that the referenced hashes do not have the possibility of malleability.

saleemrashid:
@davecgh That signature hash doesn't make sense to me. You're signing the witness hash but the witness hash contains the signature script. Or do I misunderstand?

davecgh:
That's correct.  You're signing the witness hash, which, as its name suggests, includes the witness(es).  Each witness is the signature script (and potentially the input value for the `SigHashAllValue` signature hash type).

saleemrashid:
@davecgh But the signature script contains the signature, so how can you sign the signature script?

davecgh:
Oh, yeah we're talking about the signing here.  I was confusing it in terms of how the final full hash that is included in the merkle root is calculated.  All of the above still holds, just the signature scripts are set to nil.  See [calcSignatureHash](https://github.com/decred/dcrd/blob/de705586b9896767efd62aa10f2b41835c950edf/txscript/script.go#L301-L430) in the code for details.

prusnak:
@saleemrashid I (rebase-)merged https://github.com/trezor/trezor-crypto/pull/118 - could you please change your patchset to use the new tip? Also could you please confirm this is ready to be reviewed/merged?

saleemrashid:
@prusnak Not sure if this is worth merging until Decred support is done. (I have a few more local commits for Decred support).

prusnak:
I'd like to merge at least the parts which revert Hasher-stuff from trezor-mcu and make trezor-mcu use it from trezor-crypto. Maybe a separate PR?

saleemrashid:
@prusnak In which case, you should be fine to review and merge this PR.

prusnak:
Thanks!

saleemrashid:
@davecgh When I use `SigHashAllValue`, I'm getting `invalid hashtype: 0x4`.

PR #263, opened 2017-12-11
Include file and line in Failure for DEBUG_LINK

saleemrashid:
This is especially useful for debugging the signing code (`Transaction has changed during signing` is not the most useful message!)

Should this be behind `DEBUG_LINK`? It shouldn't break anything (as the tests use the `FailureType`). If not behind `DEBUG_LINK`, we could add a `DEBUG_FSM` or similar flag (`DEBUG_LOG` is definitely not the right place for this)


prusnak:
:+1:

PR #265, opened 2017-12-12
Storage rework

jhoenicke:
- Don't copy storage from Flash to RAM.
- Use accessor functions for storage.
- New update logic to only update changed fields.


prusnak:
Thanks!

PR #266, opened 2017-12-13
Added U2F root key to storage.

### LINKED ISSUE #251: Cache the U2F private key in storage
Since U2F can be used without the PIN, we should cache the U2F private key so that we do not perform operations on the mnemonic until the device is unlocked.

jhoenicke:
Fixes #251.

Depends on trezor/trezor-common#64

### DIFF
--- a/firmware/storage.c
+++ b/firmware/storage.c
@@ -41,6 +41,7 @@
 #include "layout2.h"
 #include "usb.h"
 #include "gettext.h"
+#include "u2f.h"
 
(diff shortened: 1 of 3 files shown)


saleemrashid:
@jhoenicke I don't think you can continue to use `hdnode_private_ckd_cached` as it will constantly invalidate the cache.

prusnak:
@saleemrashid fixed in 2c63d51580b9a594a6e0bda26e9f81fd8d14a115

PR #268, opened 2017-12-13
Emulator support

saleemrashid:
Emulator support with SDL2 or headless mode, persistent flash and UDP transport.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,23 @@ env:
     - DEBUG_LINK=0 FASTFLASH=1
     - DEBUG_LINK=1 FASTFLASH=1
 
+matrix:
+  include:
+    - addons:
+        apt:
+          packages:
+            - gcc-multilib
+      env:
+        - EMULATOR=1 HEADLESS=1
+        - DEBUG_LINK=1
+      before_script:
+        - sed -i '/hidapi/d' "vendor/python-trezor/requirements.txt"
+        - pip install --user --requirement "vendor/python-trezor/requirements.txt" rlp pytest
+        - pip install --user --no-deps "vendor/python-trezor"
+      script:
+        - script/cibuild
+        - script/test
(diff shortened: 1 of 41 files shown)


saleemrashid:
Rebased. Device tests pass in 36 seconds on my laptop which is about 12 times faster than on a physical device.

saleemrashid:
> Does it make sense to move `firmware/udp.c` to `emulator/udp.c`?

While it might make sensible to move some of it (for use in non-firmware parts), some of it should remain in `firmware/` because I would like code from `emulator/` to not depend on anything from `firmware/` (in this case `firmware/usb.h`)

saleemrashid:
@prusnak I think this is an acceptable solution

.travis.yml
```diff
@@ -19,6 +19,23 @@ env:
...
+  include:
+    - addons:
+        apt:
+          packages:
+            - gcc-multilib
+      env:
+        - EMULATOR=1 HEADLESS=1
+        - DEBUG_LINK=1
+      before_script:
+        - sed -i '/hidapi/d' "vendor/python-trezor/requirements.txt"
```
prusnak: python-trezor contains `--disable-hidapi` option in `setup.py`, not sure how we can utilize this when installing via pip, though

saleemrashid: @prusnak I tried that at least 5 different ways and gave up. :laughing: `pip` will not pass provided options to `./setup.py egg_info` which is what it uses to gather dependencies.

prusnak: Ok, glad to know I am the only one who failed on this. Thanks :-)

saleemrashid: @prusnak There is an issue here but I don't imagine it will be fixed https://github.com/pypa/pip/issues/4383

prusnak:
:heart:

PR #270, opened 2017-12-15
Old storage size improved

prusnak:


prusnak:
Thanks Saleem and Jochen!

PR #271, opened 2017-12-15
bootloader: don't restore storage from unofficial firmware

prusnak:
Let's not restore storage if the previous firmware was not signed

### DIFF
--- a/bootloader/usb.c
+++ b/bootloader/usb.c
@@ -42,6 +42,7 @@
 #define ENDPOINT_ADDRESS_OUT        (0x01)
 
 static bool brand_new_firmware;
+static bool old_was_unsigned;
 
 static const struct usb_device_descriptor dev_descr = {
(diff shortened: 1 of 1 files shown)


bootloader/usb.c
```diff
@@ -437,6 +438,8 @@ static void hid_rx_callback(usbd_device *dev, uint8_t ep)
 				} while (!button.YesUp && !button.NoUp);
 			}
 			if (brand_new_firmware || button.YesUp) {
+				// check whether current firmware is signed
+				old_was_unsigned = !signatures_ok(NULL);
```
jhoenicke: Why backup metadata if old firmware was unsigned?  

This is not really a security issue in this case, but why keep sensitive information in memory if it is never needed.

prusnak:
Fixed in eb0574bda7cd4b859c5ae8728ff9c16695088b5e

PR #273, opened 2017-12-17
Adding and using typesafe MIN and MAX macros

karel-3d:
This uses simpler and safer macros for MIN and MAX

(cherrypicked from @saleemrashid code)

### DIFF
--- a/firmware/u2f.c
+++ b/firmware/u2f.c
@@ -40,8 +40,6 @@
 #include "u2f_knownapps.h"
 #include "u2f.h"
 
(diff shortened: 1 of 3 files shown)


karel-3d:
Seems like the travis tests are failing because of old GCC :(

karel-3d:
We can either use `typeof` instead of `__auto_type` or upgrade the gcc

prusnak:
Closed via 83a34ff92572f211d56f21798aa172d72a9a22dd

PR #275, opened 2017-12-18
Decred logo

peterzen:
Added Decred logo


saleemrashid:
Why? The user can set a custom homescreen if they wish.

prusnak:
The logo will be added to web interface, where user can set it as background.

PR #277, opened 2017-12-18
Update Nanopb to 0.3.9

saleemrashid:
* Stop using Nanopb for `Storage` (I have verified this generates an identical `struct` using `pahole`)
* Update Nanopb to 0.3.9
* Use Python 3 on Travis CI

All device tests pass (except for Decred, for obvious reasons).


saleemrashid:
What the hell? Is GitHub going to refuse to run a Travis CI build because of merge conflicts? https://docs.travis-ci.com/user/pull-requests/#My-Pull-Request-isn%E2%80%99t-being-built

saleemrashid:
@prusnak Success! https://travis-ci.org/saleemrashid/trezor-mcu/jobs/318319249

prusnak:
Sweeet! Can you please rebase this on top of master?

saleemrashid:
I will do that tomorrow (there are a few small changes I want to make to this pull request first). But I'm wondering if I should revert your last commit (and add the submodule back). I think it is a good idea to keep a specific version of `python-trezor` and I think that using a submodule is better than using `curl` in `script/test `.

prusnak:
I agree about using a specific version of `python-trezor` once we do a stable release of what we currently have in master. Things are in flux these days, and it would be a PITA to keep adding extra commits to trezor-mcu just because something changed in python-trezor.

(we use the same 'trick' in trezor-core, and it's okay-ish once you know about the caveats).

Once we have a next-gen stable release of python-trezor on pypi, we can install that and be done with it ...

saleemrashid:
The optimal solution would be to ship the device tests with the PyPI package, such that they are usable with `pytest`.

saleemrashid:
@prusnak I have implemented it here https://github.com/trezor/python-trezor/pull/179

firmware/Makefile
```diff
@@ -93,7 +94,7 @@ DEBUG_LINK ?= 0
 DEBUG_LOG  ?= 0
 
 CFLAGS += -Wno-sequence-point
-CFLAGS += -Iprotob -DPB_FIELD_16BIT=1
+CFLAGS += -I../vendor/nanopb -Iprotob -DPB_FIELD_32BIT=1
```
prusnak: Why do we want this change?

saleemrashid: The `-I../vendor/nanopb` is because we now need `pb_common.[ch]` and I didn't want to maintain the symlinks.

Now, the `-DPB_FIELD_32BIT=1` is to use `uint32_t` rather than `uint16_t` for the `size` member of a `bytes` field. Because `sizeof(uint32_t) == sizeof(size_t)`, we get identical behaviour to before.

I added this change because seems that, when we use a `uint16_t`, it triggers some GCC 4.8 bug. I say it's a bug because the warning does not make sense for that line of code (there is no comparison) and GCC 7.2 does not show this warning. But, while I found many GCC bugs with that warning, I could not find one similar to what we are seeing here. https://travis-ci.org/saleemrashid/trezor-mcu/jobs/318312962#L919

So, I'm not sure what to do about this.

saleemrashid: Actually, I have realized where the bug is. We are comparing `op_return_data.size < 0x100000` when that is always true because `UINT16_MAX < 0x100000`. I will revert this change and fix this warning

prusnak:
Splendid!

saleemrashid:
@prusnak I'm enjoying the evolution of your post-merge comments from "Thanks!" to "Splendid!" 😜

prusnak:
It's not evolution, Splendid is reserved for great achievements :D

PR #285, opened 2018-01-04
Cashaddr

jhoenicke:
Added support for cashaddr addresses.

Note: this makes cashaddr the default and it is no longer possible to view the old 1.. or 3...(multisig) addresses.   It is still possible to send to the old 1.. addresses, though.

Requires the corresponding pull requests for trezor/trezor-crypto#120, trezor/trezor-common#72 and trezor/python-trezor#195.


jhoenicke:
I rebased it against current master and added a third patch: Don't show coin prefix on the display.  It is implicit (the display already indicates the coin type) and it is also covered by the checksum.

jhoenicke:
This needs to be done outside the firmware for cashaddr support:

Webwallet:
- compute cashaddr addresses from xpub.  Note that only the last step from hashed public key to address needs to be changed.  The webwallet checks that the address the Trezor returns is as expected.  This check should also allow 1.. addresses so that it works with older firmware (so we don't have to deploy both at the same time)
- allow cashaddr as send to address.  The firmware supports both and both use SPENDADDRESS.  The only difference is the confirmation message given to the user, the transaction format did not change at all.

Insight-Backend:
- display cashaddr and allow searching.  Looks like blockdozer.com already implemented this (can we get their patches?)
- In principle it would also work unchanged (if the webwalllet only searches for old address format), but then the users would be confused if they go to transaction details.

Note that cashaddr only  changes the "names" for the addresses.  The transaction format is still the same.  There is a 1:1 mapping from cashaddr to legacy bitcoin address.  

Note that cashaddr differs slightly from bech32:
- longer checksum: 8 * chars instead of 6 chars (each char has 5 bit).
- Prefix ends with ':' and its checksum is computed differently.
- "Version field" is encoded differently and has other meanings.

jhoenicke:
I rebased to current master.  Had to fix the last changes to `fsm.c` to adapt to the recent `fsm_layoutAddress`

prusnak:
Rebased and merged in 059555039cf5d0960047c7b0e57b4a28b0f44be2

PR #287, opened 2018-01-06
Abort GetAddress on Initialize or Cancel

### LINKED ISSUE #247: GetAddress: receiving Initialize just shows the QR Code
... because there is no way how to distinguish user cancel (left button press) and protocol cancel (Initialize or Cancel message) returned from `protectButton()`. Needs refactor.

saleemrashid:
Fixes #247

### DIFF
--- a/firmware/fsm.c
+++ b/firmware/fsm.c
@@ -210,6 +210,23 @@ static HDNode *fsm_getDerivedNode(const char *curve, const uint32_t *address_n, 
 	return &node;
 }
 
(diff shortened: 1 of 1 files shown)


saleemrashid:
Not entirely sure in what scenario the issue described in 07f6e495b7b22c3259456c55ff19706cb5d7e1be would happen (because #247 means that `fsm_msgInitialize` would never be called anyway) but I've noticed that the change is missing from `NEMGetAddress`.

prusnak:
LGTM, Thanks!

Issue #289, opened 2018-01-15
Bech32 p2wsh addresses don't fit on display


karel-3d:
Bech32 pwsh addresses currently don't fit on display in most cases.

In examples I tried, it came 2 or 3 letters short.

Examples for example [in BIP173 itself](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#examples)

The problem will maybe be worse in the new bcash address format, if we will implement it

prusnak:
Please test

karel-3d:
Nope, this makes the last line fit, but I realized even the *previous lines* don't fit, since they are longer than the line. (I can't make picture because my phone cam is broken :( )

Issue #286, opened 2018-01-06
Display longer messages and split on word boundaries


saleemrashid:
While this is a nice-to-have in general, this is essential for supporting BIP70. With BitPay, you will get a memo that begins with `Payment request for BitPay invoice XXXXXXXXXXXXXXXXXXXXXX for merchant `. With `split_message`, that is more than 4 lines and that is not even including the merchant name.

One solution would be to having line-by-line scrolling.

 * On the initial page, the buttons would be `Cancel` and `Scroll down`.
 * On the following pages, the buttons would be `Scroll up` and `Scroll down` which scroll by a single line.
 * On the final page, the buttons would be `Scroll up` and `Confirm`.

Another issue is that we should implement a function that tries to split on word boundaries where possible.

prusnak:
Will be fixed via porting trezor-core to T1.

PR #290, opened 2018-01-15
Pack layout if lines too long

karel-3d:
This actually fixes https://github.com/trezor/trezor-mcu/issues/289

Note that with this fix, https://github.com/trezor/trezor-mcu/commit/f70772fb58126cdca155d2d5a335c605c47b93a6 is no longer needed

### DIFF
--- a/layout.c
+++ b/layout.c
@@ -32,6 +32,12 @@ void layoutDialog(const BITMAP *icon, const char *btnNo, const char *btnYes, con
 	}
 	if (line1) oledDrawString(left, 0 * 9, line1);
 	if (line2) oledDrawString(left, 1 * 9, line2);
+
+	bool packed = strlen(line3) > 21 || strlen(line4) > 21 || strlen(line5) > 21 || strlen(line6) > 21;
+	if (packed) {
+		left = 0;
+	}
+
 	if (line3) oledDrawString(left, 2 * 9, line3);
 	if (line4) oledDrawString(left, 3 * 9, line4);
 	if (desc) {


jhoenicke:
The test for > 21 is quite optimistic.  It sometimes runs out of the display already with 19 characters.

One problem is still the variable font size.  If the addresses contain too many Ms and Ws they extend outside the display.  I wonder if we should include a fixed size font, where every letter has a width of 5+1 pixels.  This would make it more predictable whether an address always fits.

BTW, bitcoincash addresses are 54 characters long (at least the ones that currently have a meaning) including the "bitcoincash:" prefix, which is part of the address, but doesn't really have to be checked by the user.  Bech32 addresses are 44 characters for simple addresses and 64 for multisig/htlc (assuming no more than 4 character prefix).  So multisig bech32 addresses are currently the bottleneck.

saleemrashid:
Related: #286 which would include wrapping on text width, not length.

karel-3d:
Width of string can be computed via `fontCharWidth`, right?

prusnak:
> I wonder if we should include a fixed size font, where every letter has a width of 5+1 pixels. This would make it more predictable whether an address always fits.

I like this. Originally, we had a non-proportional font of exactly this size, but we decided to use the proportional font for addresses too as it had better results (more letters per line) in overall.

With 6 pixels per character we can fit 21 letters per line. With 3 lines per address we can have at most 63 characters, so we are still missing one character for Bech32 multisig! :-/

mariodian:
How about the font size would get smaller as the text can no longer fit on the screen? A Bech32 multisig address is off the screen by a single character so in order to fit it in, the font size wouldn't even be so drastically smaller (just 4+1 instead of 5+1).

Have a look at [this example](http://jquery-textfill.github.io). Apple has been using it on iPhones for ages so I'd say people are even used to it. 

Another solution could be just to have 3 dots in the middle of the address. 

`bc1pw508d6qejxtdg4y5r3zarvary0c5...kw508d6qejxtdg4y5r3zarvary0c5xw7k7grplx`

It doesn't necessarily have to cause any issues since it's safe to assume people only check the beginning and the end of the address to see whether it's correct.

prusnak:
Obsoleted by https://github.com/trezor/trezor-mcu/pull/295

PR #292, opened 2018-01-20
WIP mBTC support

### LINKED ISSUE #54: Display mBTC/bits
Title says it all

saleemrashid:
Needs `Storage` field. It could be implemented two ways:

1. Always use either BTC or mBTC (depending on the user's preference)
2. Use either BTC or mBTC depending on the value (if the user has enabled this feature)

Fixes #54

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -34,7 +34,22 @@
 #include "nem2.h"
 #include "gettext.h"
 
-#define BITCOIN_DIVISIBILITY (8)
+static size_t format_amount(const CoinInfo *coin, uint64_t amount, char *str_out, size_t size) {
+	const char *suffix;
+	int decimals;
+
+	// TODO: Implement Storage field
+	if (0) {
+		suffix = " ";
+		decimals = 8;
+	} else {
+		suffix = " m";
+		decimals = 5;
+	}
+
+	bn_format_uint64(amount, NULL, suffix, decimals, 0, false, str_out, size);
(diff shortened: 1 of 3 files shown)


firmware/layout2.c
```diff
@@ -34,7 +34,22 @@
 #include "nem2.h"
 #include "gettext.h"
 
-#define BITCOIN_DIVISIBILITY (8)
+static size_t format_amount(const CoinInfo *coin, uint64_t amount, char *str_out, size_t size) {
+	const char *suffix;
+	int decimals;
+
+	// TODO: Implement Storage field
```
saleemrashid: ☝️

saleemrashid:
We could also add another field to `CoinInfo` to support bits (uBTC), lites (uLTC), atoms (uDCR), etc. and fallback to the prefix `u` if we don't have a customized name.

prusnak:
Option 1 makes much more sense than option 2.

prusnak:
Also I would like to see 4 options (BTC, mBTC, uBTC, Satoshi), not 2 (BTC, mBTC).

saleemrashid:
Not sure about the usefulness of satoshis (especially if we're not changing the units depending on the value).

prusnak:
I agree that Satoshi unit is probably not that useful, but it seems that Lightning people love that unit (judging by screenshots of lightning-app).

prusnak:
Let's not merge this. We might switch to satoshis soon :)

PR #294, opened 2018-01-30
enhace description of multisig label

farmerwampum:
On small screen I was confused what "msig" said.  Thought it was "message"
Should be plenty of room


prusnak:
Yes, this was shortened prior we reworked the layout. There is more space now, so we can apply this change.

However, change of the string is not enough. You need to increase the capacity of `desc` from `16` to `20` to accomodate for 4 extra chars and also change the assignments below from desc[5], desc[6], desc[11], desc[12] to desc[9], desc[10], desc[15], desc[16], because of the extra 4 chars you added to the beginning of the string.

PR #295, opened 2018-02-06
Support for fixed-width font

jhoenicke:
- Added a fixed width font (strokes are thinner than the default font, to fit m and w)
- Added support for multiple fonts.

~~Still missing: Send dialog should use the fixed font.~~

See also #290


jhoenicke:
bc-address
ltc-address

gen/fonts.c
```diff
@@ -1,268 +1,18 @@
...
+	{
+#include"font.inc"
+	},
+	{
+#include"fontfixed.inc"
+	},
 };
 
-int fontCharWidth(char c) {
-	return font_data[(int)(c)][0];
```
jhoenicke: BTW `(int)(c)` is broken, since char is signed.  However, `oledConvertChar` in oled.c made sure that there was no array out of bounds.

karel-3d:
1 and l looks pretty similar. I am not sure if it can be a problem - base58 doesn't have both, not sure about bech32

karel-3d:
That is probably not solvable - maybe by using serif-less small L? But that might decrease readability

jhoenicke:
Is this better?
Send Dialog

karel-3d:
> Is this better?

amazing :D

xbach:
The new lowercase "l" looks great

prusnak:
Merged via 8806accb2e3d2ac5ea97056bd81b220cdc661eb3

Thanks!

PR #297, opened 2018-02-12
change to support pi zero

heneault:
Here are my changes to run to trezor code on the raspberry pi zero device. More information can be found at http://www.pitrezor.com


saleemrashid:
Hey!

This is awesome work but

 * If you want to use the emulator for an actual device, I strongly recommend you use https://github.com/trezor/trezor-core instead. That firmware was designed with the emulator in mind, whereas this was an afterthought for tests and development (and it probably has quite a few issues. I know for a fact that the string functions I implemented aren't spec-compliant and I need to fix that)

 * It would be preferable to keep all the Raspberry Pi changes out-of-tree. Instead, you could create your own emulator library `path/to/pizero/libemulator.a` and build with `LDFLAGS=-Lpath/to/pizero`.

heneault:
Thanks for the comment!

For the moment I prefer to stay with the trezor-mcu code instead of trezor-core to match the actual device. I think the trezor-mcu code is already mostly independent of the hardware platform and the "emulator" is already a step to leverage that. I see the "emulator" more as a "linux" or "standalone" layer.

The raspberry pi code is scopped in the emulator folder to a very small section of the the button.c, oled.c and setup.c files. All the other files like rng, timer, strl... remain the sames. As you suggested, one solution could be to copy the emulator folder to call it raspberrypi with my modifications in it but this will result in a lot of code duplication between the two folders. Note that in the pi implementation I don't make use of udp.c but provide my own usb.c code which is generic for Linux (via libusbgx), thus it could be used by trezor outside of raspberry pi project. The oled_driver.c is probably the real only specific file for my project.

Another approach would be to add a raspberry pi layer on top of the emulator code but that would be to tricky I think. Emulator layer code would still need some hook points where the pi code could be called.

In all case, the Makefile.include would still need to know about that raspberrypi layer. I already did a pass to scope the change in one section only.

For the string strlcat and strlcpy implementation, you are probably aware of this but you can refer to the bsd implementation http://web.mit.edu/freebsd/head/sys/libkern/ 

Let me know in which direction you want to go!

saleemrashid:
> In all case, the Makefile.include would still need to know about that raspberrypi layer.

No, you can use `-Lpath/to/pizero` to override the emulator library. You just need to provide the same API as the emulator library. You could include, for example, the existing `emulator/flash.c` in your Pi Zero library but provide your own `emulator/oled.c` and `emulator/usb.c`.

> For the string strlcat and strlcpy implementation

Fixed in #298 

> Let me know in which direction you want to go!

I still strongly believe that we should keep other emulator libraries out-of-tree. If anything, if you make it out-of-tree, it will help others who want to implement their own emulator library for it.

But it's up to @prusnak as to what he wants to do.

prusnak:
I agree with Saleem. The current changeset is just too big and I prefer to have these emulator changes isolated in a separate directory tree.

PR #299, opened 2018-02-13
Switch to unprivileged execution and prevent access to MPU

saleemrashid:
All tests pass.

Attempts to write to the System Control Space cause a hard fault (perhaps there's a more specific fault handler we can enable for this?)


prusnak:
👍

PR #301, opened 2018-02-18
Fixing emulator makefile

karel-3d:
Fixing makefile for non-headless emulator.

Taken from @jhoenicke PR here https://github.com/trezor/trezor-mcu/pull/295 (thx @saleemrashid for tip)

for some reason, git reversed which part of Makefile is actually moved :))

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -91,8 +91,8 @@ CFLAGS   += -DHEADLESS=1
 else
 CFLAGS   += -DHEADLESS=0
 
-CFLAGS   += $(shell pkg-config --cflags sdl2)
-LDLIBS   += $(shell pkg-config --libs sdl2)
+CFLAGS   += -I/usr/include/SDL2 -D_REENTRANT
+LDLIBS   += -lSDL2
 endif
 
 else
--- a/Makefile.include
+++ b/Makefile.include
@@ -78,6 +78,14 @@ LDFLAGS  += -L$(TOP_DIR) \
 ifeq ($(EMULATOR),1)
 CFLAGS   += -DEMULATOR=1
 
+CFLAGS   += -include $(TOP_DIR)emulator/emulator.h
+CFLAGS   += -include stdio.h
+
+LDFLAGS  += -L$(TOP_DIR)emulator
+
+LDLIBS   += -ltrezor -lemulator
+LIBDEPS  += $(TOP_DIR)/libtrezor.a $(TOP_DIR)emulator/libemulator.a
+
 ifeq ($(HEADLESS),1)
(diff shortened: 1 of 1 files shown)


Makefile.include
```diff
@@ -78,22 +78,23 @@ LDFLAGS  += -L$(TOP_DIR) \
...
 
 LDLIBS   += -ltrezor -lemulator
 LIBDEPS  += $(TOP_DIR)/libtrezor.a $(TOP_DIR)emulator/libemulator.a
+
+ifeq ($(HEADLESS),1)
+CFLAGS   += -DHEADLESS=1
+else
+CFLAGS   += -DHEADLESS=0
+
+CFLAGS   += $(shell sdl2-config --cflags)
```
saleemrashid: NACK at the moment, because it breaks the build on my setup 😆 

TREZOR Core uses `-LSDL2 -LSDL2_Image` and it seems to work for everyone, so maybe we should switch to that for now

https://github.com/trezor/trezor-core/blob/30396f9cf03be20a6198ad1f76139e97c9fcc664/SConscript.unix#L97

karel-3d: Oh I did not want to leave this in, left it by mistake. (That's why git is showing it differently.)

Removing it for now

prusnak:
What's the purpose of second commit?

karel-3d:
The first comment is necessary, because it changes the order of linker flags. libemulator needs libsdl2.

Without the first commit, the flags are, on my PC,

`-lSDL2 -ltrezor -lemulator -L/home/dev/trezor-mcu/ -g3 -ggdb3 -m32  -L/home/dev/trezor-mcu/emulator`

and on my PC, that creates linker  errors like

```
/home/dev/trezor-mcu/emulator/libemulator.a(buttons.o): In function `buttonRead':                      
/home/dev/trezor-mcu/emulator/buttons.c:30: undefined reference to `SDL_GetKeyboardState'              
/home/dev/trezor-mcu/emulator/libemulator.a(oled.o): In function `oledInit':                           
/homedev/trezor-mcu/emulator/oled.c:36: undefined reference to `SDL_Init'                             
```

Reversing the order fixed it. (I copied that from @jhoenicke PR)

The second commit is necessary, because on debian-based systems (for example in the docker), `pkg-config sdl2` doesn't work and `sdl2-config` works, but on arch (and maybe others?), the reverse is true.

So per @saleemrashid tip I added the output of the command directly

prusnak:
OK. Another option would be to fix pkg-config call by using `PKG_CONFIG_LIBDIR=/usr/lib/i386-linux-gnu/pkgconfig`, but then it is again distro specific, so let's use it as you proposed. Thanks.

PR #302, opened 2018-02-18
Add TREZOR_OLED_SCALE variable to emulator

saleemrashid:
Allows scaling the emulator window. Useful for displays with high DPI.

Requested by @karel-3d :smile:

PR #303, opened 2018-02-18
Add docker build script 

karel-3d:
I have added emulator build to docker.

Note that it requires changes from https://github.com/trezor/trezor-mcu/pull/301 

The readme mentions env variable from this PR https://github.com/trezor/trezor-mcu/pull/302

### DIFF
--- a/Dockerfile.emulator
+++ b/Dockerfile.emulator
@@ -1,0 +1,17 @@
+# initialize from the image
+
+FROM debian:9
+
+# install build tools and dependencies
+
+RUN dpkg --add-architecture i386 && \
+    apt-get update && \
+    apt-get install -y \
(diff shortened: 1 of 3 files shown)


prusnak:
TBH I don't really like the change. If you really want to have this I suggest creating `Dockerfile.emulator` with the new contents and update `build-emulator.sh` to use that file (just update `IMAGE` variable to `trezor-mcu-build-emulator` or sth similar and add `-f Dockerfile.emulator` to the `docker build` call in `build-emulator.sh`).

karel-3d:
Good idea

karel-3d:
Is this better?

prusnak:
Much better ❤️

karel-3d:
It seems I am building libopencm3 which is not needed. And I install ARM packages that are not necessary. Should I make a separate PR for that?

prusnak:
Yes, please.

PR #306, opened 2018-02-20
Less paranoid change outputs.

jhoenicke:
- Allow change address to be on the receive chain (..../0/x) (see spesmilo/electrum#3920).
- Allow more than one output to the Trezor, but don't treat it as change.


prusnak:
Thank you!

PR #307, opened 2018-02-21
increase MAX_CHAIN_ID

chr1sn0:
increase MAX_CHAIN_ID to full uint32_t to fit with Pirl's Chain ID (3125659152)


prusnak:
Pirl has chosen a stupid value.

The chain_id is used to compute V value (uint32) of the signature using the following formula:

msg_tx_request.signature_v = v + 2 * chain_id + 35;

So we cannot apply this change.

PR #311, opened 2018-03-02
Indicate own dest address in send dialog

jhoenicke:
If the destination address is controlled by the TREZOR (the wallet set
the address_n field), show the path to the address on the confirm output
dialog in the same format as the "show on Trezor" dialog indicates the
path.

This is useful when claiming coins or sending from legacy to segwit wallet (provided the tool supports it). 

[The patch is a bit larger, because I had to move the `address_n_str` function to the beginning. I didn't change anything in there.]

### DIFF
--- a/firmware/layout2.c
+++ b/firmware/layout2.c
@@ -36,6 +36,122 @@
 
 #define BITCOIN_DIVISIBILITY (8)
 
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks!

PR #313, opened 2018-03-08
Wanchain

tyrion70:
Changes for Wanchain as made in trezor-core and python-trezor


karel-3d:
Are there plans for chain_id in wanchain? Any eta?

tyrion70:
Hi Karel,

As far as I know Wanchain is not planning to change the chain_id. The chain_id was kept at 1 and an extra field (txtype) was added. I'm not part of the core team however so I don't know why. (I'm not even sure if you can actually just change a chain_id in a running blockchain?). Is that a dependency for this PR then? With the txtype and the change in the slip it should be different enough from ethereum right?

thx,
Peter

prusnak:
I am disappointed, earlier you claimed you'll tell the developers to change the chain_id, so we can drop the txtype hack: https://github.com/trezor/trezor-common/pull/83#issuecomment-370454631

Now you claim something completely different ...

tyrion70:
1603 Ingalls Street, San Francisco, Californië 94124, Verenigde Staten


Hi Pavol,

My apologies, I honestly forgot about that remark because it was in trezor-common and I later solved it by adding the  txtype hack. 

 I’ve been in contact with the Wanchain team about this a few times but I neglected to update you on the matter. 

I did ask them them to change it asap, but I initially got a reply back that it might not be so easy for them to do so. 

When Karel asked for an eta I replied and again asked the team again if and when they’ll change it. I replied to Karel the way I did because I simply don’t know yet if and when they will change it and was wondering if that is a blocking issue for you. Seeing that remark again basically answers that question I guess? You want the chain-id to be different before merging this PR right?

Sorry for the long story but in short:
- The request to change is it still outstanding
- I’ll try to escalate this and get them to reply here as well with an eta

I hope this diffuses your dissapointment, it was an mistake on my side to not see the implications of your remark. 
 
Cheers, 
Peter

prusnak:
Merged, thanks!

PR #316, opened 2018-03-15
Cleaner flash handling using FLASH_PTR

jhoenicke:
Use `FLASH_PTR` macro to convert a flash address to a const pointer.
For real hardware it is just a cast, for emulator we subtract the `FLASH_ORIGIN` and use it as index into the memory mapped flash file.

This makes the emulator run with 64-bit builds as it avoids direct casts from 32 bit numbers to pointers.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -13,7 +13,7 @@ AS       := as
 
 OPTFLAGS ?= -O3
 DBGFLAGS ?= -g3 -ggdb3
-CPUFLAGS ?= -m32
+CPUFLAGS ?=
 FPUFLAGS ?=
 else
 PREFIX   ?= arm-none-eabi-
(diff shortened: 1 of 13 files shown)


saleemrashid:
utACK

If you get a chance, could you test with Clang? `make CC=clang` (or `CC=clang script/cibuild`) _should_ be sufficient.

jhoenicke:
I assume CC=clang and EMULATOR=1.  Compiles and runs with clang, but I can't do the unit tests (debuglink doesn't work for udp at the moment).  At least installing a seed and getting an address doesn't crash.

I did the unit tests for the real hardware, but some tests fail with unexpected messages.  I guess they were recently adapted for Trezor T and this breaks some Trezor 1 tests.

prusnak:
Thanks! (I accidentally rebased the branch in your remote repo, sorry).

PR #318, opened 2018-03-20
Fix size of msg_tiny and add static_assert.

jhoenicke:


jhoenicke:
On second glance, 128 instead of 136 should be enough.

prusnak:
Applied as https://github.com/trezor/trezor-mcu/commit/d6f41dba9e41ba49780878dadf157939d2eeb101

PR #322, opened 2018-03-23
Fix compilation problems & disable fastflash

jhoenicke:


saleemrashid:
Fastflash is still useful for testing bootloader changes.

emulator/setup.c
```diff
@@ -48,6 +48,10 @@ void setup(void) {
 	setup_flash();
 }
 
+void __attribute__((noreturn)) shutdown(void) {
+	for(;;);
```
saleemrashid: `for(;;) pause();` seems more appropriate here.

jhoenicke:
@saleemrashid But fastflash can't work with newer bootloaders, right?  It assumes running in privileged mode and that is not available for unsigned firmware.  Of course, you can keep a Trezor with a 1.3.4 bootloader for testing other bootloaders.

emulator/setup.c
jhoenicke: Fixed that.  But in the long run, we may think of reworking also emulatorPoll, to not hog the CPU in normal mode.

saleemrashid:
My mistake, I didn't think about that :)

prusnak:
Thanks!

PR #323, opened 2018-03-24
Enable supervisor calls

jhoenicke:
Protect flash controller using MPU.
Add supervisor calls for flash and timer.

Builds on top of my previous flashptr pull request.

Unit tests work.  Bootloader works.  Signed firmware works. Unsigned firmware works with the a small change to trezor.c `main`.


jhoenicke:
This still doesn't work well when privileged code is only in bootloader (unsigned app).

- Bootloader's system_millis is at an address the firmware may use for other purposes.
- Bootloader's __stack_chk_guard is no longer valid, but the interrupt handlers still use it.

Any ideas how to fix this?  Should we reserve some of the SRAM for the bootloader?

jhoenicke:
Unsigned firmware works.  For this, you need to comment out the setup routines in trezor.c, everything in main up to and including `mpu_config`.

prusnak:
Merged as 7fc58eb5db904f5f2de2873ab1d5e35ca157ad7e

Bertrand256:
Hello. From your comments it seems that with the commit 7fc58eb the problem with unability to execute an unsigned firmware shout be resolved, but unfortunately in my case it's not. I'm still getting the "Hardfault detected" error.

What I did:
- checked out the 7fc58eb commit
- made the changes to the trezor.c suggested by @jhoenicke 
- built a firmware and uploaded to it into two different devices
- after booting up, both of them show "Hardfault detected" error

These are lines in trezor.c that I've commented out:
`int main(void)
{
//#ifndef APPVER
//      setup();
//      __stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
//      oledInit();
//#else
//      check_bootloader();
//      setupApp();
//      __stack_chk_guard = random32(); // this supports compiler provided unpredictable stack protection checks
//#endif
//
//      timer_init();
//
//#ifdef APPVER
//      // enable MPU (Memory Protection Unit)
//      mpu_config();
//#endif

#if DEBUG_LINK
        oledSetDebugLink(1);
        storage_wipe();
#endif

        oledDrawBitmap(40, 0, &bmp_logo64);`

Am I doing it wrong or I misunderstood that it can be fixed that way?

prusnak:
@Bertrand256 Are you still using official bootloader 1.4.0? If yes, what you see is expected and you'd need to wait until we release a new bootloader as a part of the next firmware update.

Bertrand256:
@prusnak Yes, but I've also tried to build firmware from the 9c9b4bf commit, which according to log messages includes v1.5.0 bootloader.
 
I'm not very familiar with how exactly Trezor firmware & bootloader work (I'm just doing a tiny change to enable Dash Testnet support), but from the lecure of the `build-firmware.sh` script I'm assuming that it also builds a bootloader, which then is included into the resulting firmware file. 
If this is not the correct approach, I would be grateful for any directions. Thanks.

BTW. Is ETA of a new official firmware already known?

PR #330, opened 2018-03-30
Emulator: allow software renderer

yura-pakhuchiy:
Makes emulator usable in containers


yura-pakhuchiy:
According to [documentation](https://wiki.libsdl.org/SDL_CreateRenderer) SDL will give priority to accelerated renderer if it is available, but will fallback to software if it is not.

prusnak:
Thanks!

PR #332, opened 2018-03-30
Fix emulator build in docker

yura-pakhuchiy:
Emulator is a 64-bit application now, so we should install 64-bit version of libSDL. Rename docker image and emulator filename to indicate 64-bitness as well.


prusnak:
:+1:

PR #333, opened 2018-04-02
[WIP] Add GRS support 

yura-pakhuchiy:
This work-in-progress PR together with https://github.com/trezor/trezor-crypto/pull/144 adds support for Groestlcoin (GRS). However ATM it breaks support for other coins. I'm opening this PR to ask how it is better to integrate GRS specific changes.

GRS uses signle SHA256 instead of double SHA256 for transaction and message signing. How it is better to check whether we should use double or signle SHA256? Should I add new field `"double_sha256_signing"` to `coins.json`. Or just check coin name with `strcmp`? Or any other way?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "trezor-crypto"]
 	path = vendor/trezor-crypto
-	url = https://github.com/trezor/trezor-crypto.git
+	url = http://192.168.1.9:4000/trezor-crypto.git
 [submodule "trezor-common"]
 	path = vendor/trezor-common
-	url = https://github.com/trezor/trezor-common.git
+	url = http://192.168.1.9:4000/trezor-common.git
 [submodule "trezor-qrenc"]
 	path = vendor/trezor-qrenc
-	url = https://github.com/trezor/trezor-qrenc.git
+	url = http://192.168.1.9:4000/trezor-qrenc.git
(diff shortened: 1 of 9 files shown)


prusnak:
I suggest to look at Decred pull request and how it uses the concept of Hashers to achieve the similar thing. https://github.com/trezor/trezor-mcu/pull/274

yura-pakhuchiy:
@prusnak Yes, I'm using Hashers in https://github.com/trezor/trezor-crypto/pull/144 to add Groestl hash. However I'm not sure how can they help in single-SHA256 vs double-SHA256 changes in this PR. Are you suggesting to add new hasher which always perform sigle SHA256 even if `hasher_Double()` was called?

#246 uses `if (coin->decred)` a lot. Can I add similar flag for single vs double SHA256?

prusnak:
@saleemrashid any ideas? I dislike the idea of adding another flag, because GRS does not bring anything new, just different hashing. And using a single hash even when `hasher_Double` is called, sounds hackish ...

saleemrashid:
@prusnak See trezor/trezor-crypto#145

yura-pakhuchiy:
No longer needed, now all required changes are in https://github.com/trezor/trezor-crypto/pull/148

gruve-p:
Thanks for the clarification @yura-pakhuchiy

PR #334, opened 2018-04-04
nem: IV is not copied before encryption (moved to trezor-crypto)

tsusanka:
The IV copy was moved to trezor-crypto
(https://github.com/trezor/trezor-crypto/pull/140) so it is not needed
in trezor-mcu anymore


prusnak:
The build fails, because `iv` is undefined.

tsusanka:
Man, I suck. Should be fixed now.

The build goes fine now, but the tests fail. The tests are not passing in general now right?

prusnak:
Yes, they fail generally, because of emulator failure. Will investigate later. Thanks!

PR #338, opened 2018-04-08
bootloader padding, build integration, check_bootloader depend on MEMORY_PROTECT

mcudev:
These updates unify the bootloader and firmware build processes while ensuring that specific bootloader and firmware tags are used to build the respective artifacts. First the bootloader is built from a specific bootloader tag, then the actual resulting bootloader binary is used as aninput to a build of a specific firmware tag.

Also, the padding of the bootloader to 32KiB (aka "alignment") is switched to using 0x00 instead of 0xFF.

I suppose a release will actually tag and build a bootloader, then someone will update the list of `known_bootloader` hashes, tag the firmware, and then rebuild with the final bootloader and firmware tags.

Also importantly, I change to guard `check_bootloader` with `MEMORY_PROTECT`, which is the same thing that guards against locking the flash sectors in the bootloader code. I think that more cleanly protects developers.


mcudev:
Added one more commit to make `mpu_config` depend on MEMORY_PROTECT as well because when running a custom firmware, I was getting faults in the firmware at https://github.com/trezor/trezor-mcu/blob/master/firmware/trezor.c#L99 and https://github.com/trezor/trezor-mcu/blob/master/firmware/trezor.c#L103 because the bootloader had already run https://github.com/trezor/trezor-mcu/blob/master/util.h#L66

prusnak:
Nice changes, thanks!

PR #339, opened 2018-04-11
startup: shutdown at end instead of loop

prusnak:
In trezor-core we call shutdown when main returns. Let's do the same here.

One question: is the proposed patch OK, or it's better to let the code fall through? (because shutdown label is directly 4 lines below)

### DIFF
--- a/startup.s
+++ b/startup.s
@@ -33,8 +33,8 @@ reset_handler:
   // enter the application code
   bl main
 
-  // loop forever if the application code returns
-  b .
+  // shutdown if the application code returns
+  b shutdown
 
   .global shutdown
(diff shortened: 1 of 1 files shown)


prusnak:
@mcudev @saleemrashid thoughts?

mcudev:
I'd leave it with the explicit branch to `shutdown` in-case something gets moved around. I don't see how `main` in the bootloader or firmware would actually return successfully, but maybe some follow-up commits would make that happen. If you simply add a `return 0` into the firmware's `main` function right now (as a quick test), you'll get a stack smashing failure.

prusnak:
Agreed with the "moving around" remark. Thanks

PR #345, opened 2018-05-05
Use symlink to trezor-common for nem_mosaics.json

saleemrashid:


prusnak:
Saleem, can you describe why this approach is better?

saleemrashid:
It's consistent with what is done for `coins.json` and having the JSON in the same directory as the script makes more sense.

tsusanka:
Fine by me. Merging. Thanks

PR #347, opened 2018-05-11
Recovery: Don't ask for confirmation on dry run.

jhoenicke:
The message is very confusing, as it is not recovering the device. Since dry-run recovery does not change anything, there is no need to ask for confirmation.


prusnak:
Thanks!

PR #369, opened 2018-06-21
Check if firmware has privileges.

jhoenicke:
Only drop privileges if firmware is running with privileges.
Don't change the bootloader if running without privileges.

With this change the firmware can run without being signed.

### DIFF
--- a/firmware/bl_check.c
+++ b/firmware/bl_check.c
@@ -53,6 +53,10 @@ void check_bootloader(void)
 		shutdown();
 	}
 
+	if (is_mode_unprivileged()) {
+		return;
+	}
+
 	if (r == 32 && 0 == memcmp(hash, bl_hash, 32)) {
 		// all OK -> done
 		return;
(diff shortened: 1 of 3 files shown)


firmware/bl_check.c
```diff
@@ -25,6 +25,7 @@
 #include "layout.h"
 #include "gettext.h"
 #include "util.h"
+#include "buttons.h"
```
prusnak: This should not be here, right?

util.h
```diff
@@ -79,6 +79,13 @@ static inline void set_mode_unprivileged(void)
 	// http://infocenter.arm.com/help/topic/com.arm.doc.dui0552a/CHDBIBGJ.html
 	__asm__ volatile("msr control, %0" :: "r" (0x1));
 }
+
+static inline bool get_mode_unprivileged(void)
```
prusnak: Maybe change to "is_mode_unprivileged"?

firmware/bl_check.c
jhoenicke: Oops, yes.  Didn't clean up properly.

prusnak:
Thanks!

PR #372, opened 2018-06-24
Stellar: use strings instead of bytes for addresses

### LINKED ISSUE #367: Stellar: modify account fields to use addresses
Already merged in trezor-common: https://github.com/trezor/trezor-common/issues/135

To be merged in trezor-core: https://github.com/trezor/trezor-core/commit/6f0d18e1fe8aa4e2d269eeb55aa087c8d294102b

Tests are in python-trezor in the [tsusanka/stellar-accounts-addresses](https://github.com/trezor/python-trezor/tree/tsusanka/stellar-accounts-addresses) branch

### LINKED ISSUE #368: Stellar operations do not check if signing process has started
The various `fsm_msgStellar*Op` functions do not check `stellar_signing`. As such, they can be called to perform operations on the private keys, even when the PIN has not been entered.

cc @zulucrypto

zulucrypto:
* Use the new protobuf messages that send accounts as strings (fixes #367)
 * `stellar_signingAbort()` now takes an error message
 * Operations now check if a transaction is being signed (fixes #368)
 * Operations now return false on error or if the user cancels
 * Stellar `fsm_*` methods now check operation result and return early if the operation fails

This PR also re-enables the stellar code in the firmware.

### DIFF
--- a/firmware/fsm_msg_stellar.h
+++ b/firmware/fsm_msg_stellar.h
@@ -96,7 +96,7 @@ void fsm_msgStellarSignTx(StellarSignTx *msg)
 
 void fsm_msgStellarCreateAccountOp(StellarCreateAccountOp *msg)
 {
-    stellar_confirmCreateAccountOp(msg);
+    if (!stellar_confirmCreateAccountOp(msg)) return;
 
     if (stellar_allOperationsConfirmed()) {
         RESP_INIT(StellarSignedTx);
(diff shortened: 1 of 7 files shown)


prusnak:
Thank you!

firmware/fsm_msg_stellar.h
```diff
@@ -96,7 +96,7 @@ void fsm_msgStellarSignTx(StellarSignTx *msg)
 
 void fsm_msgStellarCreateAccountOp(StellarCreateAccountOp *msg)
 {
-    stellar_confirmCreateAccountOp(msg);
+    if (!stellar_confirmCreateAccountOp(msg)) return;
 
     if (stellar_allOperationsConfirmed()) {
```
saleemrashid: Could you consolidate this `if` statement (which is repeated in all the `fsm_msg*Op`) into a `stellar_requestOp` or similar function?

zulucrypto: Will do!

PR #375, opened 2018-06-25
Refactor Docker build

### LINKED ISSUE #373: Update build.sh to use script/cibuild
to consolidate the build process

saleemrashid:
Refactor Docker build to use `script/fullbuild <bootloader_commit> <firmware_commit>`, which builds in `build/{bootloader,firmware}`, copying artifacts to `build/bootloader-bl*.{bin,elf}` and `build/firmware-v*.{bin,elf}`.

Because this script uses `git rev-parse` internally, you can do tricks like `./build.sh origin/bootloader-fix upstream/firmware-wip` (as a replacement for the `REPOSITORY` parameter). Then, `git describe` will still format a nice artifact filename.

Also, the Docker build will now drop privileges to UID/GID of the source repository, so there aren't random `root`-owned files lying around in the repository.

Rewrite firmware fingerprint script from `build.sh` to `script/fingerprint`.

Haven't touched the emulator Docker build. Do we actually need this? AFAICT, it builds the emulator with dynamic linking, so it isn't particularly useful. cc @karel-3d

Fixes #373

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,10 @@ RUN apt-get update && \
 
 ENV PROTOBUF_VERSION=3.4.0
 RUN curl -LO "https://github.com/google/protobuf/releases/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-linux-x86_64.zip"
-RUN unzip "protoc-${PROTOBUF_VERSION}-linux-x86_64.zip" -d /usr
+
+# use zipfile module to extract files world-readable
+RUN python3 -m zipfile -e "protoc-${PROTOBUF_VERSION}-linux-x86_64.zip" /usr/local && chmod 755 /usr/local/bin/protoc
+
 RUN pip3 install "protobuf==${PROTOBUF_VERSION}" ecdsa
 
 RUN ln -s python3 /usr/bin/python
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,12 @@ OBJS += gen/bitmaps.o
 OBJS += gen/fonts.o
 
 libtrezor.a: $(OBJS)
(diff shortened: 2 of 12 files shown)


saleemrashid:
`./build.sh v1.6.2{,}` with this pull request will output a `build/bootloader-bl1.5.0.bin` and `build/firmware-v1.6.2.bin` that matches [releases.json](https://github.com/trezor/webwallet-data/blob/d336c7605e8edf804a5c5218f8a93806809fce47/firmware/1/releases.json)

saleemrashid:
Added `script/toolchain-download` (downloads and verifies toolchain unless it's already downloaded) and `script/toolchain-run` (checks for toolchain and executes command with toolchain in `PATH`).

So, if you don't want to use the Docker build, you can get an identical build with

```bash
script/toolchain-download && script/toolchain-run script/fullbuild bl1.5.0 v1.6.2
```

Also, cleaned up the `make` output. Commands aren't emitted unless `V=1` is provided.

```
  DEP     oled.d
  DEP     layout.d
  DEP     buttons.d
  AS      startup.o
  CC      timer.o
  CC      gen/bitmaps.o
  CC      gen/fonts.o
  AR      libtrezor.a
```

karel-3d:
I added the emulator build because I didn't know how to properly build the emulator locally - I had some problems with protoc (what `RUN unzip "protoc-${PROTOBUF_VERSION}-linux-x86_64.zip" -d /usr` does).

So I imagine more people will have the same problem (I always battle with that) and I though it's easier to just put it in Docker.

prusnak:
Rebased and merged via 7949911748019d9769d61afd935258f421f92871

PR #381, opened 2018-07-15
fixed rlp_length for chain_id > 255

hackmod:
See also https://github.com/trezor/connect/issues/137

### DIFF
--- a/firmware/ethereum.c
+++ b/firmware/ethereum.c
@@ -568,7 +568,11 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
 		rlp_length += rlp_calculate_length(1, tx_type);
 	}
 	if (chain_id) {
-		rlp_length += rlp_calculate_length(1, chain_id);
+		int length = 0;
+		if (msg->has_chain_id) {
+			length = chain_id < 0x80 ? 0: chain_id < 0x100 ? 1: chain_id < 0x10000 ? 2: chain_id < 0x1000000 ? 3 : 4;
+		}
+		rlp_length += rlp_calculate_length(length, chain_id);
 		rlp_length += rlp_calculate_length(0, 0);
 		rlp_length += rlp_calculate_length(0, 0);
 	}
--- a/firmware/ethereum.c
+++ b/firmware/ethereum.c
@@ -568,10 +568,7 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
 		rlp_length += rlp_calculate_length(1, tx_type);
 	}
 	if (chain_id) {
-		int length = 0;
-		if (msg->has_chain_id) {
-			length = chain_id < 0x80 ? 0: chain_id < 0x100 ? 1: chain_id < 0x10000 ? 2: chain_id < 0x1000000 ? 3 : 4;
-		}
+		int length = chain_id < 0x100 ? 1: chain_id < 0x10000 ? 2: chain_id < 0x1000000 ? 3 : 4;
 		rlp_length += rlp_calculate_length(length, chain_id);
 		rlp_length += rlp_calculate_length(0, 0);
 		rlp_length += rlp_calculate_length(0, 0);


prusnak:
Can you provide test vectors for various values of chain_id? Your fix is wrong because it fixes the values of chain_id for <65536 but again fails for values higher than this. Having the test vectors would find this bug, that's why I encourage you to come up with some.

hackmod:
> Your fix is wrong because it fixes the values of chain_id for <65536 but again fails for values higher than this

Yes. you are right. It is a quick hack and it fix only for `chainid < 65536` cases.

hackmod:
OK.
I've just made a modified signtx.html script to make a test.js for some selected chainId.
https://gist.github.com/hackmod/9da4963eeef8f7652691f847819b5c58

I will make some modified [bip39](https://github.com/iancoleman/bip39) tool + above script to check all possible chainId cases more easily as soon as possible.

> Using these we can create a similar test to this one, but which will check all possible chain_id values: https://github.com/trezor/python-trezor/blob/2a5ca12924d33062db2e8c0163f49568083379ea/trezorlib/tests/device_tests/test_msg_ethereum_signtx.py#L97-L117

akx20000a:
@hackmod 이런 의미의 테스트 스크립트가 아니라 빌드 테스트 내 테스트 스크립트를 말하는것 같은데요;;

hackmod:
알죠. 일단 해놓은 것 올린 것입니다.

prusnak:
I guess you are mixing chain_id and BIP44 coin_type in your tests. It is not needed to change the coin_type in BIP32 path in order to perform sign tests with various chain_id-s.

hackmod:
> I guess you are mixing chain_id and BIP44 coin_type in your tests. 
> It is not needed to change the coin_type in BIP32 path in order to perform sign tests with various chain_id-s.

yes. you are right, but anyway this testsuit is quick and dirty but handy to invest `chainid` bug right now.

hackmod:
updated to support `chainid <= MAX_CHAIN_ID (2147483630)`

image

prusnak:
@axic @Arachnid Can you comment on this issue? The fix seems rather peculiar.

eosclassicteam:
peculiar 😂  😂  😂  😂

axic:
Without having read the entire thread, shouldn't the length calculation for `chain_id` work like how `rlp_encode_number` works (serialising to big endian and finding the first non-zero byte to sense the length)?

eosclassicteam:
@prusnak Is there any TBD date for updating https://github.com/trezor/connect trezor connect api for new coins?? We are seeing "Unknown Coin" for months now 😂

eosclassicteam:
I think it is good now without patching the bug since it will teach newbie developers to build the testnet before implementing eip-155 on the mainnet haha

eosclassicteam:
Oh wait, ESN has their testnet haha

hackmod:
~~~js
const assert = require('assert');
const RLP = require('./index.js');

var ii = [0, 1, 127, 128, 255, 256, 65535, 65536, 16777215, 16777216,2147483630, 2147483631];

var kk = [];
for (var m = 1; m < 16777216; m++) {
  kk.push(m);
}

// ii.forEach(function(j) {
kk.forEach(function(j) {
  var x = RLP.encode(j);
  //console.log(j, x);
  l = j > 16777215 ? 4: j > 65535 ? 3: j > 255 ? 2: j > 127 ? 1 : 0;
  k = j > 0xffffff ? 4: j > 0xffff ? 3: j > 0xff ? 2: j > 0x7f ? 1 : 0;
  m = j < 0x80 ? 0: j < 0x100 ? 1: j < 0x10000 ? 2: j < 0x1000000 ? 3 : 4;
  //console.log('l=', l, 'k=', k, 'm=', m, x.length);
  if (l != k || k != m) {
    console.log('XXXXXXXX');
  }
});
~~~

firmware/ethereum.c
```diff
@@ -568,7 +568,11 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
 		rlp_length += rlp_calculate_length(1, tx_type);
 	}
 	if (chain_id) {
-		rlp_length += rlp_calculate_length(1, chain_id);
+		int length = 0;
+		if (msg->has_chain_id) {
```
jhoenicke: This is already in an `if (chain_id)`.  I think this nested `if` isn't necessary or even hurting.

firmware/ethereum.c
```diff
@@ -568,7 +568,11 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
...
+			length = chain_id < 0x80 ? 0: chain_id < 0x100 ? 1: chain_id < 0x10000 ? 2: chain_id < 0x1000000 ? 3 : 4;
```
jhoenicke: length should be 1 even for `chain_id < 0x80`.  `rlp_calculate_length` will take care of the special case.

````
    length = chain_id < 0x100 ? 1 : chain_id < 0x10000 ? 2 : chain_id < 0x1000000 ? 3 : 4;
````

File: firmware/ethereum.c [hunk 3]
```diff
@@ -568,7 +568,11 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
...
+		}
+		rlp_length += rlp_calculate_length(length, chain_id);
```
jhoenicke: Theoretically the second parameter should be the first byte of chain_id, i.e.  `chain_id >> (8*(length-1))`.  However practically, the first byte is only used if length is 1, so this works.

PR #386, opened 2018-07-23
support full 32bit chain_id, cast (uint64_t)v

hackmod:
with the following PR https://github.com/trezor/trezor-common/pull/174 (64bit `signature_v`)
`r,s` value can be calculated successfully.

I'm not sure why `v` value is truncated to 32bit. but anyway,
r,s values are correctly obtained.

simply recalculate `v` value in the javascript level, signing could be done successfully with full ranged 32bit chain_id.

### DIFF
--- a/firmware/ethereum.c
+++ b/firmware/ethereum.c
@@ -36,9 +36,6 @@
 #include "memzero.h"
 #include "messages.pb.h"
 
-/* maximum supported chain id.  v must fit in an uint32_t. */
-#define MAX_CHAIN_ID 2147483630
-
 static bool ethereum_signing = false;
 static uint32_t data_total, data_left;
 static EthereumTxRequest msg_tx_request;
@@ -203,7 +200,7 @@ static void send_signature(void)
 
 	msg_tx_request.has_signature_v = true;
 	if (chain_id) {
-		msg_tx_request.signature_v = v + 2 * chain_id + 35;
+		msg_tx_request.signature_v = (uint64_t)v + 2 * chain_id + 35;
 	} else {
 		msg_tx_request.signature_v = v + 27;
 	}
@@ -461,7 +458,7 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node)
 
 	/* eip-155 chain id */
 	if (msg->has_chain_id) {
-		if (msg->chain_id < 1 || msg->chain_id > MAX_CHAIN_ID) {
+		if (msg->chain_id < 1) {
 			fsm_sendFailure(Failure_FailureType_Failure_DataError, _("Chain Id out of bounds"));
 			ethereum_signing_abort();
 			return;


hackmod:
https://github.com/trezor/trezor-common/pull/174

hackmod:
image

prusnak:
Sorry, I am against this change. I think that 2 billions chain_ids are enough for everyone and using higher values than 2147483630 might break lots of other software.

hackmod:
- currently, go-ethereum(geth) does not restrict chain_id less then `2147483630` at all. this is defacto standard. Trezor simply break it
- Pirl (https://pirl.io ) use chainid `3125659152` about 11 month ago after their EIP155 hardfork.

It is more resonable for Trezor to support full range 32-bit chainId rather than the etheruem EIP155 specific encoding.

so it should be considered as a **bug** I think.

prusnak:
That's unfortunate, but I am not fixing this.

hackmod:
just to be clear, current status of official firmware.
- only chainId < 255 supported. (official firmware)
- recently, this bug found and fixed.

how about to fix `send_signature()` like as following.
~~~diff
diff --git a/firmware/ethereum.c b/firmware/ethereum.c
index a5cf7a3..a3c1689 100644
--- a/firmware/ethereum.c
+++ b/firmware/ethereum.c
@@ -203,7 +203,11 @@ static void send_signature(void)

        msg_tx_request.has_signature_v = true;
        if (chain_id) {
-               msg_tx_request.signature_v = v + 2 * chain_id + 35;
+               if (chain_id > 255) {
+                       msg_tx_request.signature_v = v;
+               } else {
+                       msg_tx_request.signature_v = v + 2 * chain_id + 35;
+               }
        } else {
                msg_tx_request.signature_v = v + 27;
        }
~~~

- chain_id < 256 cases: normal cases - return `v(signature bit) + 2 * chain_id + 35`
- chain_id > 255 cases: simply return `v`(signature bit)
  - no need to additional fix.
  - additional fix needed at the javascript level.

prusnak:
Or even shorter - no firmware change, use `1` as `chain_id`, then increase returned `v` by `(2 * chain_id - 1)`.

hackmod:
See also Ledger case https://github.com/LedgerHQ/ledgerjs/issues/168

prusnak:
> See also Ledger case LedgerHQ/ledgerjs#168

Another reason why not use chain_id wider than 32 bits

hackmod:
finally Ledger support full ranged 32bits chain_id.
- https://github.com/LedgerHQ/blue-app-eth/commit/8260268b0214810872dabd154b476f5bb859aac0
- https://github.com/LedgerHQ/blue-app-eth/commit/74c085ca2d032179fcf205cda1dee38c92877933

Ledger even returns only `1-byte` (truncated) signature `v`, but we know, `v` value contains signature `v` bit, so we can recalculate `v` at the client side, by `v = chainId * 2 + 35` + `signature v bit(0 or 1)`
already mentioned at commit https://github.com/LedgerHQ/blue-app-eth/commit/8260268b0214810872dabd154b476f5bb859aac0

> Supports 32 bits chainId for signature and matching - the returned V will be wrong and has to be recomputed by the client

for example, https://github.com/kvhnuke/etherwallet/pull/1979

PR #389, opened 2018-07-30
generate coin defs with new tooling

matejcik:
this depends on https://github.com/trezor/trezor-common/pull/173
currently the trezor-common submodule is pointed to my branch, that should be updated to master before this is merged


matejcik:
image
:)

prusnak:
It seems that Travis build fails. Otherwise I am OK with the change

matejcik:
Ready for review.
Travis failures seem to be caused only by missing Stellar.

You might want to review the "flattened protobuf symbols" commit, it's shortening the symbols after changes to nanopb.

Maybe we should hold off on accepting this until I open a conversation with nanopb upstream?

prusnak:
> Maybe we should hold off on accepting this until I open a conversation with nanopb upstream?

ACK. It will also give us some time until I merge the `next` branch (Stellar+WebUSB) back into master.

matejcik:
nanopb PR: https://github.com/nanopb/nanopb/pull/333

firmware/Makefile
```diff
@@ -131,6 +131,16 @@ clean::
 	rm -f $(1).h $(3)
 endef
 
-$(eval $(call GENERATE_CODE,coin_info,defs,coin_info.c))
-$(eval $(call GENERATE_CODE,nem_mosaics,defs/nem/nem_mosaics.json,nem_mosaics.c))
 $(eval $(call GENERATE_CODE,bl_data,../bootloader/bootloader.bin))
+
+define MAKO
```
saleemrashid: This looks overcomplicated. Since there is only one output file, it should be sufficient to use something to the effect of

```make
%: %.mako defs
    $(Q)$(PYTHON) ../vendor/trezor-common/tools/cointool.py render $<
```

matejcik: Indeed, good catch! I got confused by the original code. Also we could throw away `GENERATE_CODE` and write the one invocation explicitly.

prusnak:
Seems that https://github.com/nanopb/nanopb/pull/333 is merged now into master. Please import current nanopb master as submodule and use this newly created feature in our tooling.

prusnak:
Rebased and merged via c6b8cc8b5c4ca916d80762cd0dce391449cb2f55

Good job! :tada:

PR #390, opened 2018-08-02
commit emulator flash modification to persistent storage

heneault:
Even if the emulator flash file is opened with the O_SYNC flag, that doesn't guarantee that data is really flush out of hardware cache to persistant layer. This PR fix this.

Ref : https://thecodeartist.blogspot.com/2012/08/hdd-filesystems-osync.html

### DIFF
--- a/emulator/flash.c
+++ b/emulator/flash.c
@@ -20,10 +20,14 @@
 #include <string.h>
 #include <assert.h>
 #include <stdbool.h>
+#include <unistd.h>
 
 #include "memory.h"
 
-void flash_lock(void) {}
+void flash_lock(void) {
+	sync();
+}
+
 void flash_unlock(void) {}
 
 void flash_clear_status_flags(void) {}
@@ -131,5 +135,6 @@ void svc_flash_erase_sector(uint16_t sector) {
 uint32_t svc_flash_lock(void) {
 	assert (!flash_locked);
 	flash_locked = true;
+	sync();
 	return 0;
 }


prusnak:
Thanks!

TheCodeArtist:
Hi... i'm the author of the above blogpost.

The `sync()` call might not achieve what you want either.

**Q1. What is/are the storage media in use here?**

For eg. in case of a SATA disk,
the method to ensure that the peripheral SATA disk drive has 
written data from its internal cache to the disk platter
is by issuing [`CMD_CACHE_FLUSH`](https://en.wikipedia.org/wiki/Disk_buffer#Cache_flushing).

If `sync()` does NOT trigger a `CMD_CACHE_FLUSH`,  
then it may not be any better than using an `fd` opened with `O_SYNC`.

----

**Q2. Is there a filesystem on the block device (or accessing the block device directly)?**

If accessing a block device directly without a filesystem in between,
then opening a `fd` with `O_SYNC | O_DIRECT` can provide better guarantee that the data written via the `fd` is immediately written to the non-volatile media (and not cached in the peripheral disk's cache)

----

> Ref : https://thecodeartist.blogspot.com/2012/08/hdd-filesystems-osync.html

The entire blogpost goes into the above nuances in detail.
For a quick summary, checkout the ***Salient observations*** section at the end of the above blogpost.

heneault:
Hi,

In my project, I'm running the trezor emulator on a raspberry pi ( https:.//www.pitrezor.com ) The storage used is a SD card (filesystem mode) . The thing is that my system is powered only via usb (as a device) thus power can be removed at any time.

I did many tests and without the explicit sync call, thus keeping only the O_SYNC flag, I have corruption most of the time on a suddenly power loss. However using a sync, when the write operation is finished, to flush the write on the SD card solved that issue.

I came across your excellent post that, I think, was explaining my situation. That is why I quoted it, your explanation is better than mine ;)

TheCodeArtist:
> I did many tests and without the explicit sync call, thus keeping only the O_SYNC flag,
> I have corruption most of the time on a suddenly power loss.
> However using a sync, when the write operation is finished,
> to flush the write on the SD card solved that issue.
Practical experience beats theory any day! :-)

- BTW, you are using a micro-sd card in a pi-zero default slot right?
- Also which filesystem were you using in your experiments?

heneault:
Yes I'm using the micro-sd card. The partition is FAT (which is probably not the best one for that case).

PR #391, opened 2018-08-02
use pkg-config to find sdl2 installation correctly.

heneault:


prusnak:
This is a revert of commit https://github.com/trezor/trezor-mcu/commit/a48dcac07d9fc33d9201ef5fc3ae68d2f1c0f7a2

Can you please explain on which configurations `pkg-config` reports different than the default values?

heneault:
I'm cross compiling the emulator to ARM. The cross compile setup (yocto) is intercepting the pkg-config call to return the proper values i.e. to use the the cross compiled libraries and not the native one.

prusnak:
@karel-3d do you remember why you did https://github.com/trezor/trezor-mcu/commit/a48dcac07d9fc33d9201ef5fc3ae68d2f1c0f7a2 ?

heneault:
any updates?

prusnak:
Thanks!

saleemrashid:
@prusnak It was related to having both 32-bit and 64-bit installations of SDL2 on the system (when the emulator required a 32-bit compiler, but `pkg-config` would use the 64-bit SDL2 instead). Nowadays, the emulator should work with a 64-bit compiler.

prusnak:
Thanks @saleemrashid

PR #396, opened 2018-08-06
Support /dev/random at compile time. Default is still /dev/urandom.

heneault:
Compiling with the variable USE_RANDOM=1 will use /dev/random as random generator. Otherwise, the default is kept to /dev/urandom.

### DIFF
--- a/Makefile.include
+++ b/Makefile.include
@@ -98,6 +98,10 @@ CFLAGS   += -I/usr/include/SDL2 -D_REENTRANT
 LDLIBS   += -lSDL2
 endif
 
+ifdef USE_RANDOM
+CFLAGS += -DUSE_RANDOM=1
+endif
+
 else
 ifdef APPVER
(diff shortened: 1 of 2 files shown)


emulator/setup.c
```diff
@@ -34,11 +34,17 @@
 
 #define EMULATOR_FLASH_FILE "emulator.img"
 
+#ifdef USE_RANDOM
```
prusnak: Replace this with:

```
#ifndef DEV_RANDOM_FILE
#define DEV_RANDOM_FILE "/dev/urandom"
#endif
```

It's more flexible

Makefile.include
```diff
@@ -98,6 +98,10 @@ CFLAGS   += -I/usr/include/SDL2 -D_REENTRANT
 LDLIBS   += -lSDL2
 endif
 
+ifdef USE_RANDOM
```
prusnak: Drop this.

prusnak:
Thank you!

PR #402, opened 2018-08-15
Added fullscreen mode for emulator.

heneault:
Setting the environment variable TREZOR_OLED_FULLSCREEN to non-zero activate that mode.


prusnak:
Thanks!

Issue #409, opened 2018-09-16
Docker build does not work (old python in debian image)


karel-3d:
Debian has old python, which doesn't know f-strings. For that reason, docker build fails with

```
File "../vendor/trezor-common/tools/cointool.py", line 153
    name = crayon(None, f"({coin['name']})", dim=True)
                                          ^
SyntaxError: invalid syntax
  DEP     nem2.d
  MAKO    ethereum_tokens.c
  File "../vendor/trezor-common/tools/cointool.py", line 153
    name = crayon(None, f"({coin['name']})", dim=True)
```

prusnak:
Fixed via 307b3a742bf3d96a6a1d79b9b9abd8d7191b8809

PR #411, opened 2018-09-16
build: Use new Debian in docker for py3.6

karel-3d:
Current stable Debian uses 3.4, we use f-strings in token definitions.

Fixes https://github.com/trezor/trezor-mcu/issues/409


karel-3d:
Buster is unstable. Not sure if we want to use it. Travis build uses ubuntu.

Issue #410, opened 2018-09-16
Docker build does not work (pipenv)


karel-3d:
Even with this issue solved, by using new Debian

https://github.com/trezor/trezor-mcu/issues/409

the docker build does not work, since it is missing all the pipenv magic

PR #412, opened 2018-09-16
build: Use pipenv in docker build

karel-3d:
Fixes https://github.com/trezor/trezor-mcu/issues/410

With https://github.com/trezor/trezor-mcu/pull/411 , the docker build now works.


Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN curl -LO "https://github.com/google/protobuf/releases/download/v${PROTOBUF_V
 # use zipfile module to extract files world-readable
 RUN python3 -m zipfile -e "protoc-${PROTOBUF_VERSION}-linux-x86_64.zip" /usr/local && chmod 755 /usr/local/bin/protoc
 
-RUN pip3 install "protobuf==${PROTOBUF_VERSION}" ecdsa
+RUN pip3 install pipenv
+
```
karel-3d: ...the PROTOBUF_VERSION disappears from here though. But it is not used in pipenv too, so I don't know

PR #422, opened 2018-10-26
update instructions on how to install custom built firmware

rc125:


prusnak:
Last released version of python-trezor is still using the old syntax, also the version from master accepts both old and new syntax, so let's keep the old syntax in the docu for now.

matejcik:
That's not true :/
The syntax depends on `Click` version, not on python-trezor version. Because there's not a strict version requirement, we have no control over which syntax is right for any given user.

With that, wellp, let's just hope that our users are smart enough to figure this out.

PR #426, opened 2018-10-26
Add python 3 warning to readme

karel-3d:


karel-3d:
Oh, but you still need to use pipenv, which solves this problem too

prusnak:
@matejcik will send another PR

PR #442, opened 2018-12-29
Add ability to enter the PIN on the device itself.

kewde:
This adds the ability to enter the PIN on the device itself, using only the two buttons.

[See it in action in this video!](https://transfer.sh/Vf5vN/VID_20181229_012611.mp4)

[ NO  |  YES ]
A non-scrambled pin matrix is shown on the screen with the first number selected.
A short click of yes will move the cursor to the right.
A short click of no will move the cursor up.
If it extends the allowed range, it will flip around and not overflow.

A medium long click (until it blinks) will select a digit for entry.
A very long click will select the current digit and finish. (4x longer than the medium click).

Issues I will fix if the concept is approved:

- The long press should wait for a YesUp (release of button), else it will cause the next dialog that comes behind it, to be automatically accepted.
- pin_index has not been checked to be < 10. If you enter 11 digits, it will overflow the char array.
- Open PR for trezor-common (https://github.com/kewde/trezor-common/commit/11d7a12c72a6781a6105407fb3b7d317bdabbc33)
- Open PR for python-trezor (https://github.com/kewde/python-trezor/commit/6cc9044061c6f93b6b828acdb6d34f82152519fd)
- Fix the submodules and their commits.

### DIFF
--- a/firmware/fsm_msg_common.h
+++ b/firmware/fsm_msg_common.h
@@ -47,6 +47,7 @@ void fsm_msgGetFeatures(const GetFeatures *msg)
 	resp->has_patch_version = true;  resp->patch_version = VERSION_PATCH;
 	resp->has_device_id = true;      strlcpy(resp->device_id, storage_uuid_str, sizeof(resp->device_id));
 	resp->has_pin_protection = true; resp->pin_protection = storage_hasPin();
+	resp->has_pin_entry_on_device = true; resp->pin_entry_on_device = storage_hasPinEntryOnDevice();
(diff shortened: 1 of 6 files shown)


kewde:
Also  I am aware of the pin_index has not been checked to be <= 10. If you enter 11 digits, it will overflow the char array.

achow101:
I would like for this feature to be included in the firmware as it would make integrating the Trezor with Bitcoin Core easier.

prusnak:
Thanks for the contribution. I have cherry-picked the changes to [`pinmatrix-device`](https://github.com/trezor/trezor-mcu/tree/pinmatrix-device) branch. We'll work on it from there. Thank you!

PR #445, opened 2019-01-09
Add sha256sum check to protobuf binaries in Dockerfile

TheCharlatan:
The docker build fails if the hash is incorrect.


prusnak:
nice!

PR #453, opened 2019-02-18
Secure mnemonic checking

andrewkozlik:
config: Check mnemonic by comparing hashes instead of the actual mnemonics, to mitigate side-channel attacks.


prusnak:
Thanks LGTM: Fixed in trezor-core: https://github.com/trezor/trezor-core/commit/0751c667c3a8125d6cea01a2e34fd6522745b05e

PR #458, opened 2019-03-04
Set Bitcoin PublicKey's node to optional

tsusanka:
To make it identical to other coins and avoid bugs such as https://github.com/trezor/trezor-mcu/commit/f40219dbb609ed468aa74db25d1295fed826e547


prusnak:
Applied in 57423c9f542a5fee83a044d540b0d0d6f803f29f

PR #460, opened 2019-03-09
Add support for creating Stakenet TPoS contracts

JonSalazar:
Adds support for handling the `SignTxInput` message, which takes an input (with the `txid:index` format) and signs it using the given address.

The serialization format for the input is equivalent to how bitcoin serializes the `COutPoint` class.

This requires the changes from [trezor-common](https://github.com/trezor/trezor-common/pull/292).

### DIFF
--- a/firmware/crypto.c
+++ b/firmware/crypto.c
@@ -33,6 +33,8 @@
 #include "segwit_addr.h"
 #include "cash_addr.h"
 
+#define SERIALIZE_TX_INPUT 36
+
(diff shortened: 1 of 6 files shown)


prusnak:
This is not how we do things. First let's discuss about adding messages in trezor-common. Then, once the protocol is approved, you can proceed with adding the code.

prusnak:
Let's continue in https://github.com/trezor/trezor-common/pull/292

PR #464, opened 2019-03-13
usb: Fix segfault in usbSleep().

andrewkozlik:
Avoid segfault in usbSleep(), usbReconnect() and usbPoll() when USB is not initialized.
I am unable to test this fix, getting "Hard fault detected" during boot, however I get the same message even without this commit.

### DIFF
--- a/firmware/usb.c
+++ b/firmware/usb.c
@@ -332,7 +332,7 @@ static void set_config(usbd_device *dev, uint16_t wValue)
 		hid_control_request);
 }
 
-static usbd_device *usbd_dev;
+static usbd_device *usbd_dev = NULL;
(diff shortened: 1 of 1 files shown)


prusnak:
Remove the following 8 lines:

https://github.com/trezor/trezor-mcu/blob/master/startup.s#L22-L30

These are here only when the new firmware is being loaded from the old bootloader. I'll have to cleanup this part in a better way. :-/

prusnak:
The code changes look fine! :+1:

andrewkozlik:
Thanks, I will retest later today.

andrewkozlik:
Test successful.

prusnak:
Great! Thanks!

